PR #1, opened 2018-03-17
Make main menu optional.

Aigeruth:
Building the site fails if the menu is not defined in the config,
because the `header.html` references `.Site.Menus.main`. It can be made
optional by using the `with` helper.


luizdepra:
Thank you!

PR #14, opened 2018-04-28
Math

Vlaaaaaaad:
Hi,

There was no support for displaying math written in Latex. This is now added with the help of MathJax.

MathJax uses an external Javascript file and as not to add unnecessary load the support for math is only enabled if the `math` param is set to `true`.

The font is not ideal( see [MathJax font support](http://docs.mathjax.org/en/latest/font-support.html)), but it's a step forward.

Preview:
screen shot 2018-04-28 at 21 49 00


luizdepra:
Thank you again.
I can refactor the theme later to use a special font with math expressions.

Issue #6, opened 2018-03-27
Removing the copyright notice


qazip:
Hello,

I am interested in using your theme for a personal website (that no one will see...) and I was wondering if the licence allows to remove the bottom copyright notice, the one that states "powered by Hugo and Coder".

luizdepra:
Hey!

Yeah, the license allows the removal. The footer copyright notice refers to the content of your website, not the theme. I don't mind if you remove the "powered by" too.

If you fork someone project (with MIT license) you need to mention the original project and its author somewhere (README.md, maybe?).

BTW, I'd appreciate if you make a PR that adds a parameter to hide the footer or change its content.

PR #20, opened 2018-05-09
Update footer to be optional

flyingP0tat0:
As discussed in #6.
Adds the ability to hide (not enabled by default) the copyright notice and/or credits. Configuration was added to README and example site.


luizdepra:
Thank you. :)

PR #23, opened 2018-05-21
Typo fixed that crashes the demo build

idvoretskyi:


mbollemeijer:
Thanks for the fix, I configured the bash on my mac not to be case sensitive, so I did not notice.
Ran also into this issue when pulling the theme on my linux machine, you beat me to it ;-) thanks.

PR #27, opened 2018-05-28
Generate correct date within archetypes

khos2ow:
With this fix, after executing `hugo new foo/bar.md`, correct `Date` (essentially `NOW`).will be generated inside content file.


luizdepra:
:+1:

PR #29, opened 2018-06-01
RTL support

khos2ow:


luizdepra:
Weird. GH didn't notified my about this PR.
I'll check it later today.

Makefile
```diff
@@ -1,15 +1,31 @@
+# LESS params
 LESS_DIR = ./static/less
-LESS_FILE = style.less
+LESS_RTL_DIR = ./static/less/rtl
```
luizdepra: Just to understand... Why separate RTL rules in another file and folder?

Makefile
```diff
@@ -1,15 +1,31 @@
...
 CSS_FILE = style.min.css
+CSS_RTL_FILE = style-rtl.min.css
 CSS_TMP_FILE = tmp.css
 
 .PHONY: clean demo build
 
 build: clean
-	lessc $(LESS_DIR)/$(LESS_FILE) > $(CSS_DIR)/$(CSS_TMP_FILE)
+	for f in "`find $(LESS_DIR) -maxdepth 1 -type f`"; do cat $$f >> $(LESS_DIR)/$(LESS_TMP_FILE); done
+	for f in "`find $(LESS_RTL_DIR) -maxdepth 1 -type f`"; do cat $$f >> $(LESS_RTL_DIR)/$(LESS_TMP_FILE); done
```
luizdepra: Now, we have just one less file in each of these folders, right?
So, I thing we don't need these two lines.

Makefile
```diff
@@ -1,15 +1,31 @@
...
+	lessc $(LESS_DIR)/$(LESS_TMP_FILE) > $(CSS_DIR)/$(CSS_TMP_FILE)
 	uglifycss $(CSS_DIR)/$(CSS_TMP_FILE) > $(CSS_DIR)/$(CSS_FILE)
 	rm -f $(CSS_DIR)/$(CSS_TMP_FILE)
+	rm -f $(LESS_DIR)/$(LESS_TMP_FILE)
+
+	# RTL
+	lessc $(LESS_RTL_DIR)/$(LESS_TMP_FILE) > $(CSS_DIR)/$(CSS_TMP_FILE)
+	uglifycss $(CSS_DIR)/$(CSS_TMP_FILE) > $(CSS_DIR)/$(CSS_RTL_FILE)
+	rm -f $(CSS_DIR)/$(CSS_TMP_FILE)
+	rm -f $(LESS_RTL_DIR)/$(LESS_TMP_FILE)
```
luizdepra: And I think we can make another separate build step just to avoid the repetition of these 4 lines.

Makefile
luizdepra: Also, we can use `@include` to add include filles with LESS.

khos2ow: The reason I created another subfolder rather than `style-rtl.less` was to give future capability if we needed to add more files. And for the same reason I adjusted build process to loop through folder's content rather than reading _one_ file.

khos2ow: Yess `@include` works too and TBH I'm ok with both approach, whatever you feel more comfortable!

Makefile
khos2ow: That's a very good idea! I'll do it when we sorted out what we want to do with files/subfolder.

luizdepra:
Hmm... I don't know. I think we can just add RTL-related classes into `style.less` and do not generate another CSS file.
But, we can split `style.less` into two or more files without a problem.

khos2ow:
Actually if we ship everything in `style.less` then everything will be shipped to final artifact and `rtl = false` in config becomes a moot point, and also the following:
```
{{ if .Site.Params.rtl}}
  <link rel="stylesheet" href="{{ "css/style-rtl.min.css" | absURL }}">
{{ end }}
```
But I kinda like the idea of having one `style{-rtl}.less` which imports other less files, but I've not done this and not sure about any potential problem with it in the long run.

luizdepra:
We'll still use `rtl` in config to add the CSS class into the body tag and replace `float-right` with `float-left` and vice-versa.
But, at least we should split the color variables into another file and import it in both style files.

khos2ow:
@luizdepra can you take another look? I think it's good now!

PR #34, opened 2018-06-17
Add inverted color scheme

### LINKED ISSUE #18: Inverted colors
First I would like to thank you for neat theme! 

Second, I would like to contribute with "inverted" color scheme which could be specifiable with config parameter.  You can check example of it [here](https://masarik.sh). Maybe a bit more grey (something like #101010), i would let that up to you. 

I came up with 2 solutions:

1. Create separate css file or split css files to colors and the rest
2. Put everything to <style> tags and edit css within template - https://discourse.gohugo.io/t/generating-dynamic-css-based-on-config-file/1279/8

Would you be up for it? Do you have any suggestions / recommendations?

janmasarik:
Sorry that it took me so long 😅. Resolves #18 

There is also 1 incomplete thing which I wasn't able to fix. Multiline code doesn't have matching background color with background color of spans. 

Any suggestions on it @luizdepra ? 
image


layouts/_default/baseof.html
```diff
@@ -33,7 +33,7 @@
     {{ .Hugo.Generator }}
   </head>
 
-  <body>
+  <body {{ if (.Site.Params.invertedColors | default false) }}class="inverted-colors"{{ end }}>
```
luizdepra: I think you don't need the `default false` here. If I'm not mistaken, undeclared parameters are always zero-value, so, also false.

static/less/style.less
```diff
@@ -3,6 +3,11 @@
 @darker-bg-color: #dcdcdc;
 @darker-fg-color: #000;
 @link-color: #3366CC;
+@inverted-bg-color: #101010;
+@inverted-fg-color: #fefefe;
+@inverted-darker-bg-color: #000;
+@inverted-darker-fg-color: #dcdcdc;
+@inverted-link-color: #1ec503;
```
luizdepra: I just approved another PR that moved color variables to `colors.less`. Please, rebase and fix it.

static/less/style.less
```diff
@@ -349,3 +354,45 @@ img {
...
+  h6 {
+    color: @inverted-fg-color;
+  }
+  code {
+    background-color: @inverted-darker-fg-color;
+    color: @fg-color;
+  }
+
+  table td, table th {
+    border: 2px solid @inverted-darker-fg-color;
```
luizdepra: Do the same as you did for `pre` here. You just need to define `border-color`.

luizdepra:
@janmasarik, which `pygmentsstyle` are you using?

static/less/style.less
janmasarik: ✅ thx

static/less/style.less
janmasarik: ✅

layouts/_default/baseof.html
janmasarik: thx for pointing it out, changed 👌

luizdepra:
I think if you change the `pygmentsstyle` to any white background color style, this will work.

Maybe, `bw`? http://yjlv.blogspot.com/2015/08/pygments-styles-gallery.html#7a6d-bw

layouts/_default/baseof.html
```diff
@@ -37,7 +37,10 @@
     {{ .Hugo.Generator }}
   </head>
 
-  <body class="{{ if .Site.Params.rtl }}rtl{{ end }}">
+  <body
+    {{ if .Site.Params.invertedColors }} class="inverted-colors" {{ end }}
+    {{ if .Site.Params.rtl }} class="rtl" {{ end }}
```
luizdepra: This way, if we use both `invertedColors` and `rtl`, the result will be a double declaration of `class=`.
I don't know how to solve this problem yet. I'll look at it later and tell here.

janmasarik:
bw
image

layouts/_default/baseof.html
janmasarik: What about this?
```html
<body class="{{ if .Site.Params.invertedColors }}inverted-colors{{ end }} {{ if .Site.Params.rtl }} rtl{{ end }}">```

Comment (luizdepra) [reply layouts/_default/baseof.html]: If both options are false, they will leave and empty class attributre. `class=""`
Not cool too.

Comment (janmasarik) [reply layouts/_default/baseof.html]: It is exactly like that currently though. Does that have other than cosmetic impact? 

We can do more branching, but it would become quite polluted imo.

Comment (luizdepra) [reply layouts/_default/baseof.html]: You can leave this part as you suggested in the comment above.

If we come up with a better option here, we can change it later.

**Comment by luizdepra** [comment]
Yeah, @janmasarik. The problem here is that the color set as background for `<code>` doesn't match with the background set with `bw`.

I thing the simplest way to solve this is to use the same bg color of `bw` into `<code>` and `<pre>`, or simply don't dont setting a bg color for `<code>`.

Probably there is a better way to solve this problem, a solution that will work with any `pygmentsstyle`. I need to investigate better.

**Comment by janmasarik** [comment]
`b2` (or anything else since it seems that `b2` isn't really a valid style) with 
```
pre {
  code {
      background-color: inherit;
      color: inherit;
  }
}
```
image

Imo it looks readable and i've pushed it into my branch, so if you agree please feel free to merge. If not, I may try to play with it a bit more 😕

luizdepra:
No problem. I just need to find a better way to work with diferente code highlight themes.
For now, I'll merge your PR into a separated branch and I'll work on it.
Thank you. :)

janmasarik:
Cool, **thank you** for an awesome theme!

PR #40, opened 2018-06-27
Add language-selector for multilingualism

wasinski:
As I said it's not pretty. A lot of code is added just to have a `|` between menus and languages. I also tried to hide current language in the selector but I wasn't able to do it.

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -1,16 +1,34 @@
 <nav class="navigation">
   <section class="container">
-    <a class="navigation-title" href="{{ .Site.BaseURL }}">
+    <a class="navigation-title" href="{{ print "/" | absLangURL }}">
       {{ .Site.Title }}
     </a>
-    {{ with .Site.Menus.main }}
     <ul class="navigation-list {{ if $.Site.Params.rtl }} float-left {{ else }} float-right {{ end }}">
-      {{ range sort . }}
-      <li class="navigation-item">
-        <a class="navigation-link" href="{{ .URL }}">{{ .Name }}</a>
-      </li>
+      {{ with .Site.Menus.main}}
+        {{ range sort . }}
+          <li class="navigation-item">
+            <a class="navigation-link" href="{{ .URL | absLangURL }}">{{ .Name }}</a>
+          </li>
+        {{ end }}
+      {{ end }}
+      {{ if .Site.IsMultiLingual }}
+        {{ $node := . }}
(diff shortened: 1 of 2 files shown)


luizdepra:
Could you povide the configuration required to use the multilingual feature?
I was unable to  figure out how to show two (or more) languages alongside the menu.

Also, could you edit the README and explain how to use the feature?

luizdepra:
And, please, add yourself into the Readme's contributors section.

layouts/partials/header.html
```diff
@@ -1,16 +1,28 @@
 <nav class="navigation">
   <section class="container">
-    <a class="navigation-title" href="{{ .Site.BaseURL }}">
+    <a class="navigation-title" href="{{ .Site.Language }}">
```
khos2ow: The `href` here can be written as `{{ print "/" | absLangURL }}`.

layouts/partials/header.html
```diff
@@ -1,16 +1,28 @@
...
       {{ .Site.Title }}
     </a>
-    {{ with .Site.Menus.main }}
+    {{ if .Site.Menus.main }}
```
khos2ow: this whole block can be done as follow, IMO it's cleaner and without the repetition of `language-selector.html`:

```
<ul class="nav navbar-nav {{ if $.Site.Params.rtl }} navbar-left {{ else }} navbar-right {{ end }}">
  {{ with .Site.Menus.main }}
    {{ range sort . }}
      <li>
        <a href="{{ .URL | absLangURL }}">{{ i18n .Name }}</a>
      </li>
    {{ end }}
  {{ end }}

  {{ if .Site.IsMultiLingual }}
    {{ $node := . }}
    {{ .Scratch.Set "separator" true }}

    {{ range .Site.Home.AllTranslations }}
      {{ $language := printf "%s" .Language }}
      {{ if ne $.Site.Params.Langcode $language }}
        {{ if $node.Scratch.Get "separator" }}
          <li>
            <span>|</span>
          </li>
          {{ $node.Scratch.Set "separator" false }}
        {{ end }}
        <li>
          <a href="{{ .Permalink }}">{{ .Language.LanguageName }}</a>
        </li>
      {{ end }}
    {{ end }}
  {{ end }}
</ul>
```
I personally don't like the separator, but I added it here as an example (and it would be cool to make that configurable).

wasinski:
@luizdepra README updated,
@khos2ow huge thanks!

luizdepra:
Thank you.

PR #41, opened 2018-06-29
Add avatar option and footer text option

rdhox:
Hey,

I did some simple modifications to allow user to display an avatar on the home page, and a text on the footer. Everything is set-up from the config.toml file. Hope it's still respect the minimalist spirit of the theme, if not, I understand ;)

I just deployed my website, you can see here how it's look : https://rdhox.io/


layouts/partials/home.html
```diff
@@ -1,5 +1,8 @@
 <section class="container centered">
   <div class="about">
+    {{ if .Site.Params.avatar }}
```
luizdepra: You don't need the `avatar` config. Use `avatarurl` here. If it is missing, it will be false.

static/less/style.less
```diff
@@ -220,6 +220,16 @@ img {
   }
 }
 
+.avatar img{
+  width: 200px;
```
luizdepra: Have you tried to use units in `rem` for `width`?

static/less/style.less
```diff
@@ -220,6 +220,16 @@ img {
...
+  height: auto;
+  border-radius: 50%;
+  border: solid @darker-fg-color 1px;
```
luizdepra: I think the avatar looks better without a border.

static/less/style.less
```diff
@@ -333,7 +343,8 @@ img {
   height: 6.0rem;
   width: 100%;
   text-align: center;
-  line-height: 6.0rem;
+  line-height: 2.0rem;
+  margin-bottom:20px;
```
luizdepra: Use `rem` here  too.

luizdepra:
Also, add yourself into the contributors section at README.md.

layouts/partials/footer.html
```diff
@@ -1,5 +1,6 @@
 <footer class="footer">
   <section class="container">
+    <p>{{ .Site.Params.footercontent }}</p>
```
khos2ow: I think it's better to use:
```
{{ with .Site.Params.footercontent }}
  <p>{{ . }}</p>
{{ end }}
```
to prevent empty `<p>` if `footercontent` is missing or empty.

layouts/partials/home.html
```diff
@@ -1,5 +1,8 @@
...
+    <div class="avatar"><img src="{{ .Site.Params.avatarurl }}" alt="avatar"></div>
```
khos2ow: Also the same here:

```
{{ with .Site.Params.avatarurl }}
...
{{ end }}

rdhox:
Thank for all your observations, I will do the corrections as soon as possible!

rdhox:
Corrections done, with the ideas of @khos2ow.

luizdepra:
Ok. :)
But there are conflicts. You need to update your fork and solve them.

Issue #38, opened 2018-06-26
Code blocks have a leading indent


korrat:
When formatting a fenced code block, the first line is indented by a small amount. This is caused by the padding of `code` elements. If the padding is removed however, inline code loses a lot of background space.

Is it possible to remove the padding from fenced code blocks only?

rdhox:
Can you just create a `custom.css` file and rewrite the style of the `code` like it fit you?

Vlaaaaaaad:
@rdhox this is an issue that is affecting everybody. I tried to fix it too, but failed. 

If someone would come up with a PR for this I would be strongly in favour of it.

rdhox:
Ho ok, sorry i didn't really understand the problem. I'll try to take a look later, i was working with the css of coder yesterday, see if I can do something, and if yes, do a PR!

rdhox:
I just create a PR about the bug.

luizdepra:
Fixed in #42. Thank you.

nysthee:
Not sure if I am doing something wrong but it looks like this issue is still present on master.

luizdepra:
I never was able to reproduce this problem. Can you provide an example as images?

nysthee:
<img width="1147" alt="screenshot 2018-11-25 22 21 07" src="https://user-images.githubusercontent.com/11569296/48984776-76c95480-f100-11e8-8e86-d3b35b6985db.png">

luizdepra:
@nysthee could you send a PR, please?

PR #42, opened 2018-06-30
remove the leading indent of code element

rdhox:
Edit of the css style of code element to make it behave in `inline-block`, that remove the leading indent, saw no other (negative) effect from this edit.


luizdepra:
Thank you.

luizdepra:
Closes #38.

PR #43, opened 2018-06-30
edit README file to make disqus config working

rdhox:
The line about disqus config was inside the Params object which was making disqus not working.


luizdepra:
Thank you. :)

PR #48, opened 2018-07-06
Feature/projects

mbollemeijer:
- Adds a projects feature

screen shot 2018-07-06 at 14 31 06
screen shot 2018-07-06 at 14 31 22

I am not a less/css magician so if someone want to improve stuff be my guest :-)


static/less/style.less
```diff
@@ -1,4 +1,8 @@
-@import "colors.less";
```
luizdepra: We got regressions all over this file. Maybe a merge messed up things.

luizdepra:
Nice!
This is one of the features I'll trying to bring to the v2 of Hugo Code.

But, we need to change the grid to work on small mobile screens. Maybe one project per line?
If you're not comfortable to do this change, I can accept this PR and do it later.

static/less/style.less
mbollemeijer: I will see if I can fix this, thought I merged correctly.

mbollemeijer:
This is what I get when I scale my browser, I must admit I have not tested it anyother way.   

screen shot 2018-07-09 at 10 02 21

If this is not what it should look like, could you provide an example so I have a reference.

static/less/style.less
mbollemeijer: This should be fixed now

mbollemeijer:
@luizdepra 
Something like this ?
screen shot 2018-07-09 at 14 47 39

luizdepra:
The left margin of the grid items is not ok. It leaves a weird small gap between the alignment of the title and the grid.
I don't how to fix this problem yet. Hope we can find a solution together.

mbollemeijer:
If someone could pickup my slack, that would be great.
Havent got the time sadly :-(

luizdepra:
No problem. I will finish this feature later.

DewaldDeJager:
@luizdepra May I pick this up? Is it just the issues raised above that need to be tended to?

luizdepra:
Yep. But this PR was opened before a major refactor. Probably is a good idea to start from zero.

luizdepra:
We should redo this change in another PR.
There no point in keeping this PR open to never merge it.

PR #49, opened 2018-07-07
better Chinese font support

yukihiratype2:
wonder if you mind add support for Chinese font.


luizdepra:
I don't mind adding support to other languages, but just stacking different fonts in the CSS isn't a good solution.
Fortunately, I'm planning to add a configuration to change fonts (see #51).

yukihiratype2:
nice

PR #63, opened 2018-07-11
Mobile menu

rdhox:
Mobile menu with pure css, using the trick of the input check box to have the click event without any JavaScript. Tell me if it does the job!


khos2ow:
W00t! That's a really cool technique! I have couple of feedbacks:

- I guess it would be nicer (and generally better UX) to show the `==` on the right side for LTR and on the left side for RTL theme
- I think name of the site should not be in the dropdown menu, only the actualy *menu items*
- I think it would be nicer for the dropdown menu to be fixed and pop over the content, not pushing it down.

luizdepra:
About the language selector, I think we should move it to the footer section.

khos2ow:
Why do we want to move the language selector to the footer when: 1) we have (kind of) enough room in the header 2) it's (kind of) as important as the other menu items ?

rdhox:
Ok I edited the menu following the ideas of @khos2ow, which were great. I'm not using the ::before selector anymore to make the rtl working, which is cleaner I think. 
For the multilingual, it seems more logic to move it out the menu for people on their mobile to have directly access to the right language without going to the menu. Maybe put the multilingual menu beside the name of the website?

wasinski:
@luizdepra don't you think that knowing that particular site is available in several languages is important enough to have that selector highlighted even though it will probably be used just once?

I feel that it will be barely visible in the footer - when I was implementing this feature I looked up few other themes and for those which had language selector in the footer I would not notice it if I weren't looking up them for exactly that reason.

If it's not too much work let's make it configurable ;)

khos2ow:
@luizdepra I kinda disagree with you, switching language is as important as any other links. Correct, the frequency of that language link to be clicked might be lower, but it should be visible and prominent enough for the viewer to *know* that there's that option. Considering that if one views a blog post which most probably the footer is outside of viewable portion of page.

@rdhox I've tested your PR and I have couple of adjustment proposition, but got super busy at work. I might send it later tonight if 5à7 would actually ends at 7! ;)

This is the way I've done it for myself, so I don't mind the language selector be on the dropdown menu
selection_056
selection_057

luizdepra:
Ok, you guys convinced me about the language selector.

About the collapsible menu, I think I broke it. ahahaha



1. We can use `☰` as the menu icon, sized 2 or 2.5rem.
2. A menu with a few more items overlaps the content in a ugly way. How can we fix that? Menu itens with solid blackground?
3. I don't why, but having more than 3 menu itens breaks the menu.
4. I can use a horizontal optional separator between the menu and the language selector.

khos2ow:
@luizdepra those are the exact fixes I wanted to propose. If I remember correctly, afk, I removed floating from label and using position absolute instead, gave it a solid white background and fixed the height issue of menu items.

rdhox:
Ok I corrected bugs, and add fancy but discreet transition. I let the config files on the example site with the multilingual options like that you don't have to recreate it all to test the menu.

static/less/style.less
```diff
@@ -194,20 +194,66 @@ img {
...
+  display: none;
+}
+
+@media only screen and (min-device-width : 320px) and (max-device-width : 768px) {
+  .btn-mobile {
+    display: block;
+    font-size: 2.5rem;
+    color: black;
+    cursor: pointer;
+    padding-top: 1.5rem;
```
khos2ow: I would suggest `margin-top` instead of `padding-top`. It will produce the same result, but if in the future someone wants to enhance the button in a way, let's say give it a border, background, etc then it won't be out of place.

layouts/partials/header.html
```diff
@@ -3,32 +3,38 @@
...
-    
-        {{ range .Site.Home.AllTranslations }}
-          {{ if ne $.Site.Language .Language }}
-            {{ if $node.Scratch.Get "separator" }}
-              <li>
-                <p>{{ .Site.Params.LangSeparator }}</p>
+        {{ if .Site.IsMultiLingual }}
+          {{ $node := . }}
+          {{ .Scratch.Set "separator" true }}
+          <li class= "mobile-menu-lang-separator">
```
khos2ow: Since you put the `langseparator` in the config, it makes sense to move this inside a block, so if someone wants to override it to be `empty` it will work. something like:
```
{{ with .Site.Params.LangSeparator }}
  <li class= "mobile-menu-lang-separator">
    <hr />
  </li>
{{ end }}
```

layouts/partials/header.html
```diff
@@ -3,32 +3,38 @@
...
+              <hr />
+          </li>
+          {{ range .Site.Home.AllTranslations }}
+            {{ if ne $.Site.Language .Language }}
+              {{ if $node.Scratch.Get "separator" }}
+                <li class="multilingual-separator">
```
khos2ow: The same for here:
```
{{ with .Site.Params.LangSeparator }}
  <li class="multilingual-separator">
    <p>{{ . }}</p>
  </li>
{{ end }}
```

static/less/style.less
```diff
@@ -194,20 +194,66 @@ img {
...
+    max-height: 100rem;
+  }
+
+  .navigation-item {
+    text-align: center;
+    position: relative;
+  }
+
+  .mobile-menu-lang-separator {
+    padding-left: 7rem;
```
khos2ow: I personally like these to be `1.5rem` (both left and right) but I'm generally ok with this and specially that CSS can be easily overriden.

static/less/style.less
```diff
@@ -194,20 +194,66 @@ img {
         margin-left: 1.0rem;
         margin-right: 1.0rem;
       }
+      @media only screen and (min-device-width : 320px) and (max-device-width : 768px) {
+        float: none !important;
```
khos2ow: I personally don't like menu items in dropdown to be `center`ed, and instead be `left` aligned for LTR, and `right` aligned for RTL.

rdhox:
All your suggestions are ok for me! I'm waiting to have some more feedback and will make the corrections ;)

static/less/style.less
luizdepra: Or `line-height: 6rem;`

static/less/style.less
```diff
@@ -194,20 +194,66 @@ img {
...
-      margin-left: -1.4rem;
-    }
+.btn-mobile {
+  display: none;
+}
+
+@media only screen and (min-device-width : 320px) and (max-device-width : 768px) {
+  .btn-mobile {
+    display: block;
+    font-size: 2.5rem;
```
luizdepra: I think `2.0rem` looks better.

static/less/style.less
luizdepra: Looks good centered to me. hehehe

static/less/style.less
luizdepra: ahahaha
I prefer a smaller line. But I can leave it as is now.

With the new SCSS pipeline it will be easier to override any style rule.

rdhox:
Ok all done, even add a little of configuration, like that you can choose between centered mobile menu items or not ^^

static/less/style.less
khos2ow: As I said, I'm fine either way. We already have `custom_css` ability and SCSS you mentioned.

layouts/partials/header.html
```diff
@@ -3,32 +3,42 @@
...
-        {{ range .Site.Home.AllTranslations }}
-          {{ if ne $.Site.Language .Language }}
-            {{ if $node.Scratch.Get "separator" }}
-              <li>
-                <p>{{ .Site.Params.LangSeparator }}</p>
+        {{ if .Site.IsMultiLingual }}
+          {{ $node := . }}
+          {{ .Scratch.Set "separator" true }}
+          {{ with .Site.Params.LangSeparator }}
+            <li class= "{{ if $.Site.Params.itemscentered }} mobile-menu-lang-separator-centered {{ else }} {{ if $.Site.Params.rtl }} mobile-menu-lang-separator-rtl {{ else }} mobile-menu-lang-separator-ltr {{ end }} {{ end }}">
```
khos2ow: Since you configured it as `centered` and **other**, why not remove `mobile-menu-lang-separator-rtl` and `mobile-menu-lang-separator-ltr` and replace them with `mobile-menu-lang-separator-full` (or something similar)
and give it equal padding left and right `1.5rem`? (still personal opinion)

rdhox: Yes no problem !

PR #65, opened 2018-07-14
Asset pipeline

### LINKED ISSUE #52: [v2] Replace LESS with SCSS
We should replace every LESS code with SCSS and use the new [Hugo Piper](https://gohugo.io/news/0.43-relnotes/) feature.

luizdepra:
I have chaged all styling code to SCSS, using the new Hugo's Asset Pipeline.

I need help testing here.

1. Test RTL feature.
2. Test using Hugo non-extended version.
3. Test trying to override SCSS rule.

Closes #52 .


khos2ow:
I might be able to test RTL out, but fist I need to understand how SCSS work in the first place!

rdhox:
I checked the modified files and i'm curious about the advantages of changing for SCSS? On the code it's not that clear. (It's not a critic at all, more for my personal knowledge ^^).

luizdepra:
@rdhox, about the language, SCSS/SASS is more used than LESS. In fact, they are quite similar.
I'm changing to SCSS to take advantage of the Hugo's Asset Pipeline. With the pipeline added, we'll not depend on node.js to build CSS files anymore. Hugo will do it for us. So, this change will benefit anyone that want to make changes in the theme style.

rdhox:
Thanks for your answer, I understand now. I will try to test some features as soon as I can, but I unfortunately broke my wrist recently... So I'm afraid i'm not gonna be a great help in a near future :/

luizdepra:
These changes seem ok so far. Merging soon.

khos2ow:
Finally I got the time to test this new Pipeline feature. It's so cool! I'm still playing around with it, and most probably will open a PR soon to move `RTL` back (since I'm using it on my own site :wink:)

luizdepra:
Well, with Asset Pipeline is possible to use RTL or inverted colors by just overriding the SCSS files.
Try to do this. I think it is simpler this way.

But, we can try the `@if` and `@else` in the SCSS approach.

I just don't want to create lots of styling options (via theme configuration) that can be achieved by overriding style files. Inverted Colors is one of these cases. You just need to override `_variables.scss` to change the colors (and soon the fonts too).

khos2ow:
That's generally true, specially for inverted colors, but most probably will not work for RTL, because most people (including myself) who eventually need this, will end up with two languages on the site one being LTR the other being RTL. and technically we cannot have _one_ final css for both, that's why I think adding a config to theme (the same as before, `rtl = true|false`) and based on that include `coder-rtl.css` (which has been compiled out of `corder-rtl.scss`) will be required.

luizdepra:
Understood. How about creating a base CSS class, `use-rtl` or something, and add it to the `body` tag. In the SCSS, we can use nested overrides.

khos2ow:
Yes, that's exactly what I had in mind, similar to what there was in v1.0

PR #72, opened 2018-08-30
Enable RTL layout by providing 'rtl = true' in the config

khos2ow:
This is to enable RTL layout. I've created `assets/scss/coder-rtl.scss` and corresponding required files, and modified `baseof.html` so that one can have a bilingual, multi-layout design in one site.


assets/scss/_colors.scss
```diff
@@ -0,0 +1,6 @@
+// Colors
+$bg-color: #fefefe !default;
+$fg-color: #323232 !default;
+$alt-bg-color: #dcdcdc !default;
+$alt-fg-color: #000 !default;
+$link-color: #3366CC !default;
```
luizdepra: This file will conflict with my styling refactor. But, don't mind, I will fix it.

assets/scss/_content.scss
```diff
@@ -46,8 +46,6 @@
         width: 20.0rem;
         text-align: right;
         margin-right: 3.0rem;
-        text-align: right;
-        margin-right: 3.0rem;
```
luizdepra: Why removing these lines?

layouts/_default/baseof.html
```diff
@@ -28,6 +28,18 @@
...
 
+    {{ if .Site.Params.rtl }}
+      {{ if .Site.IsServer }}
+        {{ $cssOpts := (dict "targetPath" "css/coder-rtl.css" "enableSourceMap" true ) }}
+        {{ $styles := resources.Get "scss/coder-rtl.scss" | resources.ExecuteAsTemplate "style.coder-rtl.css" . | toCSS $cssOpts }}
+        <link rel="stylesheet" href="{{ $styles.Permalink }}" media="screen">
+      {{ else }}
+        {{ $cssOpts := (dict "targetPath" "css/coder-rtl.css" ) }}
+        {{ $styles := resources.Get "scss/coder-rtl.scss" | resources.ExecuteAsTemplate "style.coder-rtl.css" . | toCSS $cssOpts | minify | fingerprint }}
+        <link rel="stylesheet" href="{{ $styles.Permalink }}" integrity="{{ $styles.Data.Integrity }}" media="screen">
```
luizdepra: Refresh me. Why do we create a separated CSS for RTL?
Why not just add it to the existing ones?

assets/scss/_content.scss
khos2ow: duplicate of the two lines above it. most probably copy/paste issue.

luizdepra: Ooh! I did not notice these 2 lines above.

assets/scss/_colors.scss
luizdepra: It is just a matter of renaming this file. Don't worry.
I created a similar file in my `style-refactor` branch. 

https://github.com/luizdepra/hugo-coder/blob/style-refactor/assets/scss/_variables.scss

khos2ow: That's really cool. I actually wanted to suggest similar thing, and also add an empty `_override.scss` file in the theme and include it in `coder.scss` which can be used by the user to _override_ from their site.

PR #74, opened 2018-09-03
Use hugo-wrapper for demo target of exampleSite

khos2ow:
hugow is a platform independent wrapper for hugo binary
and it removes the need to actually have the correct
binary installed, rather than controlling the required
version through a version file checked in the repo.


luizdepra:
Great project!

I think we can change 3 things here.
1. **Don't add hugow directly into the repository. Create a Makefile rule to download it.** This way any user can easily update hugow version without changing any theme files. 
2. **Don't force a specific version of Hugo, just get the newst extended one by default.** It is better to let the user define which version to use, but by default we should get the newest one.
3. **If `hugo` is present in `${PATH}`, use it instead of `hugow`. Or make `hugow` optional.** I think some users (me included) will like to use a global installation of hugo.

khos2ow:
Thanks for the feedback. But IMO all your points contradict the whole purpose of `hugow`. And also note that since this repository is a _theme_ and will be included in actual website repo somewhere else there will be two sides of `hugow`:

1. in this repo, for the purpose of demo of `exampleSite`
2. in some other repo, which users who see the guide and `exampleSite` here might get the idea and do similar thing about `hugow` on their own repositories.

That being said:
1. as you can see `hugow` is in `exampleSite` folder and only being used by `demo` target of Makefile and has nothing to do with actual usage of theme on an actual site. So it's up to the user to include it in their own repository which uses `hugo-coder`.
2. Unfortunately I still don't have a solution for this yet. If you don't check in `.hugo/version` file in your repository `hugow` will always download the **latest normal** binary. But on the other hand maintaining one file (`version`) to have universally control of which Hugo version and type (extended or not), to me, will not be a big deal.
3. Why? :) They are both literally the same thing. With `hugow` the script will automatically download the same thing that you manually download it yourself (either `tar` or `deb`). And also how do you solve for portability on different machines (one user multiple machines, or multiple users collaborating, or even CI machines)? That was the main point I created `hugow` in the first place.

luizdepra:
I kinda use exampleSite to develop this theme. I'm afraid that hugow can, in some way. change mu workflow. But, let's try first and fix later.

khos2ow:
I really don't think it will cause any issue, on the other hand it will help in terms of ease of upgrading/downgrading Hugo or even test the site for particular version of Hugo.

Still all these are possible to manually downloading and installing Hugo (`deb` or `tar`) but this will encapsulate that in `version` file, which should be easier to manage.

PR #77, opened 2018-09-07
Style refactor

luizdepra:
Well, this PR will change drastically how this theme looks like... ok, not so much.

Here I'm fixing readability with these points:

- Better Fonts: changing from FireCode to Lato + Merriweather.
- Fixing Typography: changing letter and paragraph spacing and sizing.
- Better Colors: changing to colors from the Material Design palette.

My site is already using these changes: https://luizdepra.com/

I will keep this PR open for a couple of days to collect feedback.

### DIFF
--- a/assets/scss/_base.scss
+++ b/assets/scss/_base.scss
@@ -13,7 +13,7 @@ body {
   display:flex;
   color: $fg-color;
   background-color: $bg-color;
-	font-family: 'Fira Mono', monospace;
+	font-family: $text-font-family;
 	font-size: 1.6em;
 	font-weight: 400;
 	letter-spacing: 0.0625em;
@@ -54,6 +54,7 @@ h3,
 h4,
 h5,
 h6 {
+  font-family: $heading-font-family;
   color: $alt-fg-color;
   text-transform: uppercase;
   letter-spacing: 0.0625em;
@@ -106,6 +107,7 @@ h6 {
 }
 
 pre {
+  font-family: $code-font-family;
   margin: 1.6rem 0 1.0rem 0;
   padding: 1.6rem;
   overflow-x: auto;
@@ -113,6 +115,7 @@ pre {
 
 code {
   display: inline-block;
+  font-family: $code-font-family;
   background-color: $alt-fg-color;
   color: $bg-color;
   padding: 0.4rem 0.8rem 0.4rem 0.8rem;
--- a/assets/scss/_base.scss
+++ b/assets/scss/_base.scss
@@ -14,19 +14,18 @@ body {
   color: $fg-color;
   background-color: $bg-color;
 	font-family: $text-font-family;
-	font-size: 1.6em;
-	font-weight: 400;
-	letter-spacing: 0.0625em;
+	font-size: 1.8em;
+	font-weight: 300;
   line-height: 1.8em;
-  @media only screen and (min-device-width : 320px) and (max-device-width : 480px) {
-    font-size: 1.4em;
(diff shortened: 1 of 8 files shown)


dalenoe:
Looks sharp. Nice job

khos2ow:
Thanks @luizdepra this looks great. I have some comments:

- Headings look bigger and out of proportion. I believe reverting back the size would make it look better. i.e. move back `h1` font-size to `3.2rem` (as opposed of current `5.4rem`) and also `line-height`, and correspondingly the others heading.
- the links in `container page` and `container post` look somehow inconsistence with the rest of the site (being red). I believe a black underscore with hover of red would look better.
  - on that note, I think it's kinda cool if we change the `page.html` and `post.html` to wrap `{{ .Content }}` in a `<div class="page-content">` or something. this way from our side we can easily target anything inside it and it gives a very good reference point for the users if they want to override the theme on their site.
- I suggest we add an empty `assets/scss/_override.scss` file and add the import in `coder.scss` to give a  reference point to the user to _override_ the theme on their site without changing anything in the `config.toml` or `baseof.html`
- I have a change, almost ready, for better code highlight, that can be either included in this PR or we merge this and handle that in another PR of its own. Otherwise I think, specially with this change of fonts` having a solid white and solid black for code fences don't look too good anymore.

luizdepra:
@khos2ow 

- Headings are in a different proportion. I followed some design best practices and fixed another proportion. I can try again with the old proportion an see if it looks ok with the new fonts, but I'm happy with the current result.
- The link is a hard decision. I can't make every link read or black. Red menus would be weird and black links with underscore don't differ from a normal text with underscore. What I can experiment is: use a grey color for links and use it on social icons, footer, text links. So, the link color will be present in every page.
 - There is no need to create another wrapper. `post`, `page` and `list` are made for this purpose.
- You don't need a empty file to do this, you can override any existing SCSS file. I can make a wiki page or something about it later.
- I think I solved the problem with code fences with this refactor. `background-color: inherit;`

luizdepra:
So, I reverted back some properties and changed the link color back to blue.

screenshot from 2018-09-11 16-01-14
screenshot from 2018-09-11 16-01-18
screenshot from 2018-09-11 16-01-23
screenshot from 2018-09-11 16-01-08

Vlaaaaaaad:
Looks good! I liked the "_hackery_" feel of the previous version but this is definitely more readable.

I would argue for links( just the ones in the post content) to be underlined too as to follow accessibility best practices( just blue colour in a dark room with flux turned on kind of renders the links invisible).

luizdepra:
@Vlaaaaaaad, the blue looks fine here in night mode. Anyway, you can always change it to any other color you want.

chipsenkbeil:
I don't care for the underscore in links in the menu, so I would have argued for the opposite of @Vlaaaaaaad where there is no underscore in the menu links. Suppose I can always add my own styling in custom css to adjust later. Looking forward to this landing.

khos2ow:
@luizdepra fair enough on your points of heading and links, but I still think having an empty override file will be beneficial. True, one can customize the file with _the same name_ on their website but that means they need to copy the whole file and keep the whole file.

For example, let's say I'm super happy with everything except the top margin of component X in `_content.scss`, if we have the `_override.scss` I can put only **one** css rule in the file and it's done, otherwise I need to copy `_content.scss` over to my site's repo and _change_ one line.

At the end they are producing the exact same result, but having an override capability will be cleaner IMO.

luizdepra:
@khos2ow, ok. We can create this file after this PR.

I'll merge it now.

PR #87, opened 2018-10-02
Add Dutch config params

sexybiggetje:
For my new layout for Ginnegappen.nl I chose this theme.
So updated exampleSite to Dutch for hacktoberfest :).


luizdepra:
Hey. Thank you for your contribution.
Well... The `config.toml` from `exampleSite` is just an example configuration used in the Hugo's Themes Showcase. It already have an extra language, so there is no need to add another one.

Anyway, thank you again.

Issue #53, opened 2018-07-09
[v2] Template refactor


luizdepra:
I think we can improve some templates, maybe splinting them in more files.
It is interesting to read Hugo's template documentation and look for improvements.
Also, I want to review all CSS rules.

Suggested:

- [x] Split files and reuse common code.
- [x] Remove extra whitespaces.
- [x] CSS Rules.
- [ ]  Better image resizing with [Hugo Image Processing](https://gohugo.io/content-management/image-processing/).
- [x]  Configurable date format.
- [x] Move `post` related templates to a separated folder.

khos2ow:
Since the scope of the theme is going beyond the about and post only the immediate suggestions I have for this one move `post`to its own standalone folder. With this move:

- any other list will not be effect posts
- it will be a working example for any other potential list/page (such as `projects`)

I've done this kind of separation in the past, it's quite easy, but I might not have time to do it until next week. So assign this to me also may be?

luizdepra:
These 3 improvements are still wanted.

khos2ow:
I completely forgot about `post` refactor. I'll send a PR soon.

PR #88, opened 2018-10-03
Post refactor to hold everything its own

khos2ow:
Partially fix issue #53


layouts/posts/single.html
```diff
@@ -2,5 +2,24 @@
...
+      <header>
+        <h1 class="title">{{ .Title }}</h1>
+        <h2 class="date">{{ .Date.Format (.Site.Params.dateFormat | default "January 2, 2006" ) }}</h2>
+      </header>
+
+      <div>
+        {{ .Content }}
+      </div>
+
+      <br />
```
luizdepra: This `<br />` is really necessary? All alignment should be done with CSS rules.

layouts/posts/li.html
```diff
@@ -0,0 +1,4 @@
+<li>
+  <span>{{ .Date.Format (.Site.Params.dateFormat | default "January 2, 2006" ) }}</span>
+  <a href="{{ .URL }}">{{ .Title }}</a>
+</li>
```
luizdepra: I don't think we should split this one if we aren't using the code in two or more places.

layouts/posts/single.html
khos2ow: Not really, it was like this before and I kept it. I'll adjust the css rules for it.

layouts/posts/li.html
khos2ow: True, it might only be used once, but that's the whole point of this. `li` here is one _view_ of the `post` type, potentially in the future we might have `summary`, `short`, `expanded`, etc as other views for post.

Essentially these views are standalone representation of that particular type.

layouts/posts/single.html
luizdepra: No problem. You can leave it to another PR if you want.

layouts/posts/li.html
luizdepra: Understood.
So, Does not make sense to move this file to `partials/post/`?

luizdepra: Also, why you used these constructions with `{-` and `-}`? Never saw it before. Can you share any documentation about it?

khos2ow: Not always, this is the difference between "view of something" and "reusable piece of code". For example there's `single` view (which is a default known one to Hugo, but a view nonetheless) is defined outside of partial on its own directory but uses partial (`disqus` and `math`) which are not _view_ of any type.

khos2ow: Those `{{-` and `-}}` and left and right white-space trimming in go template ([here](https://golang.org/pkg/text/template/#hdr-Text_and_spaces)).

Go processes everything in `{{` and `}}` and retains white space and carriage return, so one line of html code inside a multi-line of go _logic_ will have multiple extra empty lines around it (such as (e.g. empty spaces around `<link hre="coder.css" />`)

It looks nicer when viewing source of the generated page without extra empty spaces, when not required.

layouts/posts/single.html
khos2ow: I updated this PR anyway with css rule instead of `<br />`.

PR #90, opened 2018-10-04
Taxonomy initial commit, tags and categories

### LINKED ISSUE #57: [v2] Tags and Categories Taxonomy
Add tags and categories into the post header. They can be placed near the post date or bellow, what looks better. 
We also can create custom list pages for both taxonomies.

khos2ow:
Taxonomy initial commit. It will show

- tags
- categories

on top of each post.

**No Tags**
image

**Tags and Categories**
image

It also shows designated title for taxonomy pages:

**Tag**
image

**Category**
image

Potentially we can show these taxonomies in `/post` list as well, which currently is not part of the PR, so any comments or suggestions are welcomed.

Fixes #57


khos2ow:
@luizdepra I would suggest to proceed with #91 first so I can rebase this on top of that to be able to use the Netlify to validate the PR.

khos2ow:
@luizdepra this should be ready to go.

khos2ow:
I agree. I put them at the top because you mentioned it in #57

> Add tags and categories into the post header. They can be placed near the post date or bellow, what looks better.

I personally like it better all the way _after_ the post body at the bottom. Otherwise you're research is totally welcome.

luizdepra:
Reading time is a nice feature to have too.

khos2ow:
@luizdepra I'm showing taxonomy both at the header and footer of the post for comparison. I personally am in favor of the footer. Whichever you decide I can follow up on it and fix the layout.

khos2ow:
@luizdepra ping.

luizdepra:
changes
I've tested and changed it a bit.
Well, I thinks is simpler if I accept your PR and make the changes later. I will do it this weekend.

khos2ow:
I enabled "allow edits from the maintainers", you ahould have access to directly work on this PR rather than merge it and fix it later.

luizdepra:
Ok. Done.
I changed the icons to black, and fixed some SCSS and HTML code.

preview_taxonomy

khos2ow:
Looks nice, thanks!

PR #91, opened 2018-10-04
Enhance Makefile and netlify deployment

### LINKED ISSUE #79: Setup deploy-preview on Netlify to test PRs
Although this is a _theme_ and not an actual website, but essentially every change we push (thorough pull requests) should be tested against `exampleSite` and we can easily setup a `deploy-preview` pipeline on Netlify to generate the example site for that particular PR at hand to be able to immediately see the change in action without the need to pull down the PR locally to test..

**But** first we need to solve for or come up with a workaround (hopefully not checking-in `resources` folder) for the fact that Netlify cannot handle Hugo extended in their CI.

khos2ow:
This PR enhances "deploy preview" on Netlify and Makefile targets.

Fixes #79


Makefile
```diff
@@ -1,17 +1,18 @@
 .PHONY: release build demo clean
```
luizdepra: You need to add `prepare` here.

khos2ow: Right, good catch!

netlify.toml
```diff
@@ -0,0 +1,6 @@
+[build]
+  publish = "demo/public"
+  command = "make prepare && hugo --source demo"
+
+[context.deploy-preview]
+  command = "make prepare && hugo --source demo --buildFuture --buildDrafts --baseURL $DEPLOY_PRIME_URL"
```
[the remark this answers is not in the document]
khos2ow: done.

luizdepra:
Thank you. I had to add some extra configuration into `netlify.toml` but everything is working fine.

PR #97, opened 2018-10-13
Crossorigin=Anonymous on theme css 

### LINKED ISSUE #96: Theme CSS missing crossorigin="anonymous"
If you run the same hugo deploy on multiple URLs like I do then you run into this problem where the theme style sheet can't be loaded. The stylesheet link tag as the integrity attribute which is great, but is missing the `crossorigin="anonymous"`. You will see the other stylesheets that have integrity have that attribute. It only really matters that it is missing when the URL of the page is not the same domain as the theme stylesheet URL, which in my case does happen. Pull request for fix on the way.

> Subresource Integrity: The resource 'https://example/css/coder.min.578f8f...8135600.css' has an integrity attribute, but the resource requires the request to be CORS enabled to check the integrity, and it is not. The resource has been blocked because the integrity cannot be enforced.

prenagha:
Add crossorigin=anonymous to the theme css links to support CORS type deployment where the baseURL may be different than the URL used to see the web page (origin). Also makes consistent with the CSS links of font-awesome and normalize above. When using integrity with a cross origin request you need the crossorigin attribute on the link otherwise the browser will not load the resource.
Fixes #96

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -25,7 +25,7 @@
     {{ else }}
       {{ $cssOpts := (dict "targetPath" "css/coder.css" ) }}
       {{ $styles := resources.Get "scss/coder.scss" | resources.ExecuteAsTemplate "style.coder.css" . | toCSS $cssOpts | minify | fingerprint }}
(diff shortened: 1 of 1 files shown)


luizdepra:
Thank you.

PR #101, opened 2018-10-14
Add CONTRIBUTORS file

khos2ow:
- Add `CONTRIBUTORS` file
- ~~Add `authors` target in `Makefile` to populate list of authors based on git history~~
- Remove the Contributors section from `README`


AUTHORS
```diff
@@ -0,0 +1,24 @@
+Chip Senkbeil <chip.senkbeil@gmail.com>
```
luizdepra: I would like to name this file as `CONTRIBUTORS.md`. We can use [this example](https://github.com/jonathantneal/svg4everybody/blob/master/CONTRIBUTORS.md) as model, but we should not list every change made in this file, just list the people.

luizdepra: Also, I prefer to list the contributors and only link to their GitHub profile. By listing their emails we create a free spam list. :P

AUTHORS
```diff
@@ -0,0 +1,24 @@
...
+Harry Khanna <harry@khanna.cc>
+Ihor Dvoretskyi <ihor@linux.com>
+Jacob Wood <jacoblukewood@gmail.com>
+Jan Baudisch <dev@baudisch.xyz>
+Jiri Hubacek <jiri.hubacek@gmail.com>
+Khosrow Moossavi <khos2ow@gmail.com>
+Luiz de Prá <luiz.pra@olist.com>
+Luiz de Prá <rawarkanis@gmail.com>
+Luiz F. A. de Prá <luizdepra@users.noreply.github.com>
+Luiz F. A. de Prá <rawarkanis@gmail.com>
```
luizdepra: I appear here 4 times. ahahahaha
But I don't need to be listed in this file.

Makefile
```diff
@@ -16,3 +16,6 @@ prepare: clean
 
 clean:
 	rm -rf demo
+
+authors:
+	git log --all --format='%aN <%aE>' | sort -u > AUTHORS
```
luizdepra: It is interesting to update the contributors list automatically. But, as I already pointed in the previous comment, there is some flaws.
I would like to create a PR template instead. This template should asks to the contributor to update the `CONTRIBUTORS.md` file.

https://embeddedartistry.com/blog/2017/8/4/a-github-pull-request-template-for-your-projects

khos2ow: The problem with manual updating the file (separate or as part of the README) is the it requires maintenance and people tend to not do it. We already have this issue as there are people who have contributed but their name is not on the list. That's why I tried to make this as automated as possible.

AUTHORS
khos2ow: To me this's fine.

AUTHORS
khos2ow: I'm ok with using `CONTRIBUTORS` as opposed to `AUTHORS`. But at the same time I don't see a point why not show their emails. That's a personal choice everyone makes, to make their emails private or not. If I set it to be public (which you can clearly and easily find it through git history), there's no reason to not show it here in this file too.

luizdepra: People may not what to add their email to this file, so I prefer to just make a list and link to their GitHub profile.

khos2ow: All right, I'm gonna change it to GH handles.

khos2ow:
@luizdepra this should be ready now.

luizdepra:
Thanks. :+1:

PR #103, opened 2018-10-16
Fixed reference to old _colors variable name

p-rintz:
Using RTL would fail due to _colors no longer being available. Changed to import _variables instead.


luizdepra:
Thanks.

PR #112, opened 2018-11-03
Add trailing slash to remove redirects.

tobinjt:
When you request /tags/foo, the web server will redirect you to
/tags/foo/ so the browser needs to make two requests.  Add the trailing
slash in the link so that only one request is needed.


crachel:
Also needs it for /categories/foo/ since both point to directories.

tobinjt:
> Also needs it for /categories/foo/ since both point to directories.

Good point, added.  Thanks :)

luizdepra:
Hmm... I need to check if there is a better way to solve this problem ratter than concatenating a `/`.

layouts/partials/taxonomy/categories.html
```diff
@@ -4,6 +4,6 @@
     {{- if gt $index 0 }}
       <span class="separator">•</span>
     {{- end }}
-    <a href="{{ ( print "categories/" ( . | urlize ) ) | relLangURL }}">{{ . }}</a>
+    <a href="{{ ( print "categories/" ( . | urlize ) "/" ) | relLangURL }}">{{ . }}</a>
```
luizdepra:
```suggestion
    <a href="{{ ( printf "categories/%s/" ( . | urlize ) ) | relLangURL }}">{{ . }}</a>
```
It is better this way.

layouts/partials/taxonomy/tags.html
```diff
@@ -4,6 +4,6 @@
     {{- if gt $index 0 }}
       <span class="separator">•</span>
     {{- end }}
-    <a href="{{ ( print "tags/" ( . | urlize ) ) | relLangURL }}">{{ . }}</a>
+    <a href="{{ ( print "tags/" ( . | urlize ) "/" ) | relLangURL }}">{{ . }}</a>
```
luizdepra:
```suggestion
    <a href="{{ ( printf "tags/%s/" ( . | urlize ) ) | relLangURL }}">{{ . }}</a>
```
Same here.

luizdepra:
@tobinjt, I just suggested two fixes.
Also, add yourself into `CONTRIBUTORS.md`.

tobinjt:
I've accepted those two fixes and added myself to CONTRIBUTORS.md, thanks!

luizdepra:
Thanks. :)

PR #114, opened 2018-11-08
Remove flexbox on body

### LINKED ISSUE #113: Remove flexbox on body
Is there any reason to add `display: flex;` onto `body`?

It's breaking the layout on Google Web Cache.

josephting:
Fixes #113.


khos2ow:
hmm, I think this broke the Avatar and logos on home page, they are not vertically aligned anymore. That was the usage of `display: flex` specially important for home page.

josephting:
Sorry, I missed that completed. However, I think that has to do with the height of `section.container.centered` instead of flexbox of body.

If we add these properties below, it will fix that problem.

```css
.content {
  position: relative;
}
.centered {
  position: absolute;
}
```

PR #117, opened 2018-11-11
Paginate By Date

mishfit:
# Pull Request Checklist
- Paginate by date by default
  - Ordinary page listings order by weight, date, link title, then file path. Pagination orders by date modified. I've added front matter date back in as a default order
- If left as is, post dates can be jumbled if modified date and "published" (front matter) date don't match. I've initialized a Paginator with options passed in to `.Paginate`
- [ ] If you have changed any SCSS code, run `make release` to regenerate all CSS files.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -22,3 +22,4 @@
 - [Vinícius dos Santos Oliveira](https://github.com/vinipsmaker)
 - [Vlad Ionescu](https://github.com/Vlaaaaaaad)
 - [Joseph Ting](https://github.com/josephting)
+- [Mish Ochu](https://github.com/mishfit)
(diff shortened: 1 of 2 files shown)


luizdepra:
I don't see the necessity to change the default behavior.
Your PR changes the date order from `newest -> oldest` to `oldest -> newest`.

mishfit:
I apparently wasn't clear enough. I think this is extremely important as [the default range paginates by weight, date, linktitle, filepath](https://gohugo.io/templates/lists/#default-weight-date-linktitle-filepath) whereas the paginator ignores the front matter date and instead uses the modified date of the markdown file.

This is important when migrating a site with old posts and you move the pages over in an arbitrary order.

All that is necessary to restore `oldest -> newest` is to add `.Reverse` to the paginator constructor.

luizdepra:
Could you provide a repository/example with this problem?

The default ordering should work well if you don't apply different weights for your pages.

PR #118, opened 2018-11-16
Fix homepage main content section vertical align

josephting:
**UPDATE**: Requesting for some feedback for the suggestion [below](https://github.com/luizdepra/hugo-coder/pull/118#issuecomment-441854448).

---

This fixes an issue found by @khos2ow after merging #114.

Adding `position: relative;` to parent element while having `position: absolute;` to the child element will ensure that the `height: 100%;` be applied relative to the parent element and not itself.

`left: 0; right: 0;` is to make sure that the width of child element span across the width of parent element.

If this looks good, let me know so I'll squash the commits before merging them.


josephting:
I found out a proper way to using flexbox that can avoid adding `position: absolute` to `.centered`.

`.wrapper { justify-content: space-between; }`

However, this will cause any page that has short content to get center aligned vertically e.g. post list.

With that said, I propose to add homepage specific class to apply the CSS rule above only to homepage.

Let me know what you guys think. @luizdepra @khos2ow 

Pushing the change with `justify-content` in a minute for preview.

luizdepra:
Yep. The current fix centers every page. https://deploy-preview-118--hugo-coder.netlify.com/posts/
I need to check for another options.

josephting:
Can confirm that the layout works fine on Firefox.

On Chrome and Safari, it looks like this.

https://deploy-preview-114--hugo-coder.netlify.com/
deploy-preview-114--hugo-coder netlify com_

khos2ow:
Sorry guys, I've been very busy in the past couple days. I personally don't think removing `display: flex;` is a good idea as using flex is a good practice now, but I'm not sure how to test #113 (i.e. how to test on `Google Web Cache`). But if removing `display: flex` is the only option to solve that, I would suggest to narrow it down with some selector:
- define a class on root page's body (e.g. `homepage`)
- remove it like:
  ```
  .homepage {
    .wrapper {
      display: block;
    }
  }
  ```

josephting:
Thanks, @khos2ow. I was actually trying to avoid having to add class to `baseof.html`.

With that, I came up with this solution that's ready for testing via the preview. Confirmed it works on Chrome and Firefox.

https://deploy-preview-118--hugo-coder.netlify.com

Let me know what you think.

luizdepra:
Sorry for taking too much time to come back here.
I completely ignored your last submission, @josephting. Sorry again.

It looks fine. I just need to test everything again. I'll try to do the tests tonight.

luizdepra:
So, I didn't find any problem with this approach. Thank you for the effort and patience.

PR #119, opened 2018-11-18
Accept boolean as well as string

Kyslik:
Some flags / switches accept only boolean value of `true` (`pygmentscodefences`), but `math` flag currently accepts only string `"true"`.

This change introduces no breaking change, math can be both `true` and `"true"`. When V2 hits it can be changed to accept only `true`.

### DIFF
--- a/layouts/partials/posts/math.html
+++ b/layouts/partials/posts/math.html
@@ -1,4 +1,4 @@
-{{- if eq .Params.math "true" -}}
+{{- if (or (eq .Params.math true) (eq .Params.math "true")) -}}
   <script type="text/javascript" async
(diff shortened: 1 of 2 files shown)


layouts/partials/posts/math.html
```diff
@@ -1,4 +1,4 @@
-{{- if eq .Params.math "true" -}}
+{{- if (or (eq .Params.math true) (eq .Params.math "true")) -}}
```
luizdepra:
```suggestion
{{- if .Params.math -}}
```
I think this is a better way to handle the problem.

luizdepra:
Also, add yourself into `CONTRIBUTORS.md`

layouts/partials/posts/math.html
Kyslik: @luizdepra of course it is better lol :D; I just assumed that `true != "true"` in the conditional context.

Thank you.

luizdepra:
TY.

PR #121, opened 2018-11-21
Change copyright year to be current

### LINKED ISSUE #109: Multilingual support of the copyright
When viewing a website in the primary language, the year in the copyright is fine, but when switching languages it becomes something like "0001".

josephting:
Fixes #109.


luizdepra:
Ty.

PR #138, opened 2019-01-15
Easy customize

Nepomuceno:
### Prerequisites

### Description
Adding classes to the listing so we can avoid this messing up customization that people might end up doing in `li.html` 

### Issues Resolved

It resolves the fact of anyone trying to configure a custom post list.

#### General

I am including the classes on lists to facilitate the customization of the template.

#### Resources

#### Contributors


assets/scss/_content.scss
```diff
@@ -71,7 +71,7 @@
       @media only screen and (max-width : 768px) {
         margin: 1.6rem 0 1.6rem 0;
       }
-      span {
+      span.date {
```
luizdepra: We can remove the `span` here.

assets/scss/_content.scss
```diff
@@ -81,7 +81,7 @@
           text-align: left;
         }
       }
-      a {
+      a.title {
```
luizdepra: Remove the `a` here too.

assets/scss/_content_inverted.scss
```diff
@@ -5,7 +5,7 @@ body.inverted {
     .list {
       ul {
         li {
-          a {
+          a.title {
```
luizdepra: Same here.

assets/scss/_content_rtl.scss
```diff
@@ -20,7 +20,7 @@ body.rtl {
   .list {
     ul {
       li {
-        span {
+        span.date {
```
luizdepra: Here too.

luizdepra:
Nice. I just thing we should remove some tags from the SCSS code, so we can use the class with other tags. I've marked the spots that need to change.

Nepomuceno:
Corrected the recommendations thank you!

luizdepra:
Thank you. :)

PR #154, opened 2019-03-08
Fix links

Lednerb:
Preview was not correct because of 2 whitespaces.


luizdepra:
Ahahaha. I missed that too. Thank you.

PR #171, opened 2019-04-16
Replace deprecated .Hugo with hugo.

tobinjt:
### Prerequisites

Potentially, I don't know for sure.

### Description

Hugo 0.55 deprecates `.Hugo` and produces this warning:

`Page's .Hugo is deprecated and will be removed in a future release.  Use the global hugo function.`

Update the theme to use `hugo` rather than `.Hugo`.

I marked this as potentially introducing a breaking change because I don't know how backwards compatible this change is.


tobinjt:
I don't understand why the tests are failing :(
The only error message I can see is `11:34:23 AM: failed during stage 'building site': Build script returned non-zero exit code: 255`
Is there a doc I can read about how to run the tests myself to debug them?

Thanks,

luizdepra:
I think the problem is the Hugo version used to build the deploy preview. You can change it in the [`netlify.toml`](https://github.com/luizdepra/hugo-coder/blob/master/netlify.toml) file.
Also, you need do change the minimal hugo version in the [`theme.toml`](https://github.com/luizdepra/hugo-coder/blob/master/theme.toml) file.

tobinjt:
Thanks for the info!
I've updated the minimum version and the version to test with to 0.54 after brute-force searching to find the minimum version supporting the new variable.
You might not want to accept this PR - it significantly increases the minimum required version.

There's another deprecation warning:
`Page's .URL is deprecated and will be removed in a future release. Use .Permalink or .RelPermalink. If what you want is the front matter URL value, use .Params.url.`
Sadly Hugo doesn't report *where* it's used, and I don't understand the intent of the theme well enough to make the change :(

luizdepra:
There is no problem if we need to bump the minimal version of this theme. It is good to keep things up-to-date.

Thank you for the PR and for pointing the deprecation warning. I will create a Issue to remember to fix this problem later.

PR #177, opened 2019-04-28
fix taxonomy & multilingual typo

yudiwdynto:
fix taxonomy & multilingual typo

### DIFF
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -43,9 +43,9 @@ disqusShortname = "yourdiscussshortname"
     custom_css = []
 
 [taxonomies]
-  categories = "categories"
-  tags = "tags"
+  category = "categories"
   series = "series"
+  tag = "tags"
 
 [[params.social]]
     name = "Github"
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -43,6 +43,8 @@ disqusShortname = "yourdiscussshortname"
     custom_css = []
 
 [taxonomies]
+  categories = "categories"
+  tags = "tags"
   series = "series"
 
 [[params.social]]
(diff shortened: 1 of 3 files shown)


exampleSite/config.toml
```diff
@@ -43,6 +43,8 @@ disqusShortname = "yourdiscussshortname"
     custom_css = []
 
 [taxonomies]
+  categories = "categories"
```
luizdepra: Are you experiencing any problems with taxonomies or translations?
Because there is no need to define these taxonomies here, they are default in Hugo.
Also, taxonomies are defined like this: `singular = "plural"`.

yudiwdynto: yes, i'm facing 404 page not found,
also translations, it just show " : Development ", i though it should be " tag: Development " or " category: Development "

exampleSite/config.toml
```diff
@@ -43,6 +43,8 @@ disqusShortname = "yourdiscussshortname"
...
+  tags = "tags"
```
luizdepra: Ok, reading the docs I found out that we need to redeclare the default taxonomies (tags and categories) when adding a custom one (series). So, you just need to change this file by adding tags and categories in the `singular = "plural"` form.

```
[taxonomies]
  category = "categories"
  series = "series"
  tag = "tags"
```

i18n/en.toml
```diff
@@ -1,7 +1,7 @@
-[category]
+[categories]
 other = "category"
 
-[tag]
+[tags]
```
luizdepra: You should revert this file.

i18n/pt-br.toml
```diff
@@ -1,7 +1,7 @@
-[category]
+[categories]
 other = "categoria"
 
-[tag]
+[tags]
```
luizdepra: This one too.

luizdepra:
Also, please fix the example `config.toml`.

luizdepra:
Greate! Ty.

PR #181, opened 2019-05-11
use safeHTML in footer

Naata:
While building website with your AWESOME theme I stubled upon a small problem.
When I tried to put link in footer i got this:
image

My small change allows user to do this:
image

I hope this modification matches your overall architecture of the theme :)


luizdepra:
Nice, this is a good improvement.
Thank you.

PR #198, opened 2019-07-14
Check if commentoUrl is set before using it.

tobinjt:
### Prerequisites

- [X] This pull request fixes a bug.

### Description

Prevents including HTML that will not work and that refers to a non-existent file.

### Issues Resolved

None.

### Checklist

#### General

- [X] Describe what changes are being made
- [X] Explain why and how the changes were necessary and implemented respectively

When a variable does not exist it evaluates to `nil`, and so the check `(not (eq .Site.Params.commentoUrl ""))` evaluates to true rather than false because `nil` is not "".  Before this change, unless you explicitly set `Params.commentoUrl` to `""`, the commento HTML block would be added, and this does not work in the browser because `/js/commento.js` does not exist.  This change ensures that the commento HTML block is only added when Params.commentoUrl exists and is set to a non-empty string.

#### Resources

#### Contributors

### DIFF
--- a/layouts/partials/posts/commento.html
+++ b/layouts/partials/posts/commento.html
@@ -1,4 +1,4 @@
-{{- if and (not (eq .Site.Params.commentoUrl "" )) (eq (.Params.disable_comments | default false) false) -}}
+{{- if and (isset .Site.Params "commentourl") (not (eq .Site.Params.commentoUrl "" )) (eq (.Params.disable_comments | default false) false) -}}
 <div id="commento"></div>
 <script src="{{ .Site.Params.commentoUrl }}/js/commento.js"></script>
 {{- end -}}
--- a/layouts/partials/posts/disqus.html
+++ b/layouts/partials/posts/disqus.html
@@ -1,3 +1,3 @@
-{{- if and (not (eq .Site.DisqusShortname "" )) (eq (.Params.disable_comments | default false) false) -}}
+{{- if and (isset .Site "disqusshortname") (not (eq .Site.DisqusShortname "" )) (eq (.Params.disable_comments | default false) false) -}}
   {{ template "_internal/disqus.html" . }}
 {{- end -}}


tobinjt:
@alrayyes because this changes your code and I'd like you to check the change is OK please.

An example of the problem is https://www.johntobin.ie/blog/probers_for_my_hosting/ - look at the source, search for commento, and you'll see this:

```
<div id="commento"></div>
<script src="/js/commento.js"></script>
```

I'll leave it like this for a few days so you can see it live and then I'll update with the changes in this PR.

Thanks,

alrayyes:
Sorry about that, that was a stupid mistake. PR works for me, tested with and without the parameter set, don't see any problems.

layouts/partials/posts/commento.html
```diff
@@ -1,4 +1,4 @@
-{{- if and (not (eq .Site.Params.commentoUrl "" )) (eq (.Params.disable_comments | default false) false) -}}
+{{- if and (isset .Site.Params "commentourl") (not (eq .Site.Params.commentoUrl "" )) (eq (.Params.disable_comments | default false) false) -}}
```
luizdepra: Maybe we should make the same change in `disqus.html` too.

alrayyes: or just use Hugos [internal diqus template](https://gohugo.io/templates/internal/#disqus):

```
{{ template "_internal/disqus.html" . }}
```

luizdepra: Yeah, we use it (here)[https://github.com/luizdepra/hugo-coder/blob/master/layouts/partials/posts/disqus.html].

tobinjt: @alrayyes We all make mistakes, it's not a big deal :)

@luizdepra I made the same change to disqus.html, but I don't see a similar change in output?  Possibly I'm just not triggering it?

luizdepra: Yep. Just to be sure.

luizdepra:
Ty guys.

PR #201, opened 2019-07-31
Add Rmarkdown (.Rmd) example file and MathJax support

ecosang:
I use coder theme for my R blog (with [blogdown](https://bookdown.org/yihui/blogdown/) package). 
Add Rmarkdown (.Rmd) example file and MathJax support from https://github.com/yihui/hugo-lithium/blob/master/layouts/partials/footer_mathjax.html  

However, I am not very familiar with javascript and Hugo. Thus, please review this if this works well. Thanks.

### DIFF
--- a/exampleSite/content/posts/2019-07-31-r-rmarkdown_test.html
+++ b/exampleSite/content/posts/2019-07-31-r-rmarkdown_test.html
@@ -1,0 +1,83 @@
+---
+title: "Hello R Markdown"
+author: "Frida Gomam"
+date: 2019-07-31T21:13:14-05:00
+categories: ["R"]
+tags: ["R Markdown", "plot", "regression"]
+output:
+  blogdown::html_page:
+    toc: true
+    fig_width: 6
+    dev: "svg"
+---
+
+
+<div id="TOC">
+<ul>
+<li><a href="#r-markdown">R Markdown</a></li>
+<li><a href="#including-plots">Including Plots</a></li>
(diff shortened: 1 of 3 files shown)


layouts/partials/posts/math.html
```diff
@@ -22,6 +22,10 @@
     });
   </script>
 {{- end -}}
+{{ if and (not .Params.disable_mathjax) (or (in (string .Content) "\\") (in (string .Content) "$")) }}
```
luizdepra: MathJax is already here. Look at the previous lines.
But, maybe we should review it.
Could you remove this part by now?

exampleSite/content/posts/2019-07-31-r-rmarkdown_test.html
```diff
@@ -0,0 +1,83 @@
...
+tags: ["R Markdown", "plot", "regression"]
+output:
+  blogdown::html_page:
+    toc: true
+    fig_width: 6
+    dev: "svg"
+---
+
+
+<div id="TOC">
```
luizdepra: I think there is no need to add this html file here.

luizdepra:
So RMarkdown works with Hugo without any other dependencies?
Is there any reference so I can learn more about it?

ecosang:
Ah, after building a Rmd file again, MathJax is not the only problem with hugo_coder. Here is the link https://github.com/rstudio/blogdown . But, I think it is better to close this for now. Thanks.

PR #223, opened 2019-10-04
Chinese Transaltion

Jain-Sanchit:
### Prerequisites

### Description

Chinese translation

### Issues Resolved

Add Chinese translation

### Checklist

#### General

- [ ] Describe what changes are being made

#### Resources

#### Contributors

### DIFF
--- a/i18n/chn.toml
+++ b/i18n/chn.toml
@@ -1,0 +1,24 @@
+[category]
+other = "类别"
+
+[tag]
+other = "标签"
+
+[series]
+other = "系列"
+
+[reading_time]
+one = "一 分钟 读"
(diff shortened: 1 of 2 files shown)


luizdepra:
Ok. You're adding two languages here.
First, Chinese which we already have [here](https://github.com/luizdepra/hugo-coder/blob/master/i18n/zh-cn.toml). What is the difference?
And then Japanese too.

You translated even the `.Count` template variable in both languages. Are you familiar with Go Templates and Hugo?
Also, the filename/code seems wrong for both languages.

Finally, these translations seems too literal. I don't mean to be rude but, are you fluent in both languages or you just translated these texts using Google Translator?

PR #230, opened 2019-10-21
Make LaTeX compatible with MathJax version 3

bobbywlindsey:
MathJax was updated to version 3 which broke the configuration blocks. `MathJax.Hub.Config()` function doesn't exists in the new version per the [mathjax docs](http://docs.mathjax.org/en/latest/upgrading/v2.html#configuration-changes).


luizdepra:
Looks good. Ty.

PR #234, opened 2019-11-07
Automatic dark/light colorschemes

### LINKED ISSUE #226: Auto-switch between Dark and Light modes
With some simple vanilla javascript, it's possible to automatically switch between Light and Dark mode, based on the user's OS preference. 

## Proposal

Add an option to `config.toml` like this:

`dark_mode = "auto"` to (by default) automatically detect the user preference.

This could replace the `inverted` option:

`dark_mode = "light"` to always select light mode
`dark_mode = "dark"` to always select dark mode

## Impact

This would remove the `inverted` option, and add a new `dark_mode` option. It might be possible to re-use the `inverted` option, but the naming is a bit off. It would also mix boolean values with a string value.

## Implementation

The implementation is straight-forward. I'm currently doing this already with some custom JavaScript at https://www.devroom.io (try switching between light and dark mode on your device).

At the moment you need to set `inverted = true` to include the inverted CSS. The javascript will then add/remove the `.inverted` class to `body`, based on what preference it detected. If no preference can be found, the script could default to light or dark, or set either one based on the local user's time. 

@luizdepra Please let me know you thoughts on this, I'd be happy to create a PR for this.

thesmiley1:
### Prerequisites

### Description

A new `colorscheme` option is added with the capability to be always dark,
always light, or automatic based on the user's browser setting.  This
makes the `inverted` option superfluous, and therefore deprecated.
Nomenclature of files, variables, etc are changed from `inverted` to
`dark` to reflect this change.

### Issues Resolved

Resolves #226.

### Checklist

#### General

#### Resources

I did so, but there is no change to generated css files.  It seems the reason for that is because dark (and rtl) options are not enabled in the demo config, so they get excluded in `baseof.html`.  Is this intentional?

#### Contributors


thesmiley1:
There are a few things worth noting:

Adding a dark option makes the inverted option superfluous and it wasn't clear to me what the compatibility expectation would be.  I chose to leave the inverted option functioning as is, but marked as deprecated.

I was unable to find a way to completely avoid duplicating styles.  If I understand correctly, it is not possible because it is not possible to combine selectors and media queries.  Sass mixins keep the source dry, but some styles are repeated in the final output.

I touched more than I had originally intended to in the spirit of embracing "dark" over "inverted", by changing file/variable/style names from "inverted" to "dark".

luizdepra:
Hey! Thank you for the PR.
I'm kinda busy right now, but I'll try to take a look at this PR late this week.

luizdepra:
Ok, @thesmiley1. This PR looks fine.
About the `inverted` config, do you mind removing it in this PR?
There is no problem if you can't do it right now. I can merge this PR and remove this config later.

thesmiley1:
Sure, I will put something together soon.

thesmiley1:
I pushed another commit to remove the `inverted` option completely and rebased to fix the merge conflict in `CONTRIBUTORS.md`.

PR #239, opened 2019-11-25
Add \\( and \\[ as mathjax delimiters

bebound:
### Prerequisites

### Description

`\(...\)` and `\[...\]` are widely used latex equation delimiters, so add them into delimiter option.

### Issues Resolved

Now mathjax and katex should be able to render `\\(x\\)` properly.

### Checklist

#### General

#### Resources

#### Contributors

### DIFF
--- a/layouts/partials/posts/math.html
+++ b/layouts/partials/posts/math.html
@@ -6,10 +6,7 @@
     MathJax = {
       tex: {
         inlineMath: [
-          ['$', '$']
-        ],
-        displayMath: [
-          ['$$', '$$']
+          ['$', '$'], ['\\(', '\\)']
         ],
         processEscapes: true,
         processEnvironments: true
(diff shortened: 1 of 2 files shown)


layouts/partials/posts/math.html
```diff
@@ -6,10 +6,7 @@
     MathJax = {
       tex: {
         inlineMath: [
-          ['$', '$']
-        ],
-        displayMath: [
-          ['$$', '$$']
+          ['$', '$'], ['\\(', '\\)']
```
luizdepra: I think we still need `displayMath` here. Shouldn't it be something like this here:
```
    inlineMath: [
      ['$', '$'], ['\\(', '\\)']
    ],
    displayMath: [
      ['$$', '$$'], ['\\[', '\\]']
    ],
```

bebound: The default option has treated `$$` and `\\[` as delimiters, so we don't need to override `displayMath`. Remove it should be fine.

luizdepra: Ok. Undestood.

luizdepra:
Thank you.

Issue #263, opened 2020-02-10
Add support for Fathom Analytics


yashmehrotra:
Fathom Analytics is a simpler and more privacy-focused alternative to Google Analytics. They are open source as well (https://github.com/usefathom/fathom)

I have tweaked this theme on my local setup and have got a working model for this. All an end user would have to do is the following lines in `config.toml`

```toml
[params.fathomAnalytics]
    siteID = "ABCDE"
    rootURL = "analytics.example.com"
```

If you are interested in such a feature, I'd be happy to create a PR for the same.

yashmehrotra:
Closing as it is fixed by #264

PR #264, opened 2020-02-10
Add support for Fathom Analytics

yashmehrotra:
### Prerequisites

### Description

Have added support for Fathom Analytics (https://usefathom.com/). Fathom is a privacy-first site analytics tool. I implemented this because hugo only provides out of the box support for Google Analytics. There are very few themes which support other analytics tools as well.

### Issues Resolved

Fix for #263 

### Checklist

#### General

#### Resources

#### Contributors

### DIFF
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -50,6 +50,11 @@ disqusShortname = "yourdiscussshortname"
     # Custom JS
     custom_js = []
 
+# If you want to use fathom(https://usefathom.com) for analytics, add this section
+[params.fathomAnalytics]
+    siteID = "ABCDE"
+    rootURL = "analytics.example.com"
+
 [taxonomies]
   category = "categories"
   series = "series"
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -53,7 +53,8 @@ disqusShortname = "yourdiscussshortname"
 # If you want to use fathom(https://usefathom.com) for analytics, add this section
 [params.fathomAnalytics]
     siteID = "ABCDE"
-    rootURL = "analytics.example.com"
+    # Default value is cdn.usefathom.com, overwrite this if you are self-hosting
+    serverURL = "analytics.example.com"
 
 [taxonomies]
   category = "categories"
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -108,6 +108,10 @@
 
     {{ template "_internal/google_analytics.html" . }}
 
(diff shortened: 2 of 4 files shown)


exampleSite/config.toml
```diff
@@ -50,6 +50,11 @@ disqusShortname = "yourdiscussshortname"
     # Custom JS
     custom_js = []
 
+# If you want to use fathom(https://usefathom.com) for analytics, add this section
+[params.fathomAnalytics]
+    siteID = "ABCDE"
+    rootURL = "analytics.example.com"
```
luizdepra: I just don't get what is `rootURL` for. It should points to the root of the site or is a link to the analytics panel?

yashmehrotra: It points to the analytics panel, fathom provides a paid service but can be self hosted as well. So, to keep it open-ended, this is being taken as a param

yashmehrotra: @luizdepra Should I reword the variable to something more understandable ?

luizdepra: Yup. Maybe serverURL or something like that.
If there is a hosted option for FathomAnalytics and its URLs will not change by user, you can provide a default value in the template.

layouts/_default/baseof.html
```diff
@@ -108,6 +108,10 @@
 
     {{ template "_internal/google_analytics.html" . }}
 
+    {{ if .Site.Params.fathomAnalytics }}
```
luizdepra: This template need the ID and the URL to work. So, it is better to also check if the ID was provided. Maybe the URL to if there is no default.

exampleSite/config.toml
yashmehrotra: Cool, will do that

layouts/_default/baseof.html
yashmehrotra: Ok, makes sense

yashmehrotra:
@luizdepra I have made the changes you asked. Please take a look

layouts/_default/baseof.html
```diff
@@ -108,6 +108,10 @@
 
     {{ template "_internal/google_analytics.html" . }}
 
+    {{ if .Site.Params.fathomAnalytics.siteID }}
```
luizdepra: Almost there.

We still need to check for `fathomAnalytics` or we will get warning when not using Fathom.
```suggestion
    {{ if and .Site.Params.fathomAnalytics .Site.Params.fathomAnalytics.siteID }}
```

yashmehrotra:
@luizdepra Done

PR #277, opened 2020-03-15
Rework commit info display in footer

gfeun:
### Prerequisites

### Description

Current commit feature requires passing commit hash environment variables during hugo build and will display the same commit hash for every page.

This adds complexity at build time because you have to add a git command to retrieve commit hash values before building the site

This pull request proposes to use the Hugo GitInfo feature (https://gohugo.io/variables/git/) to get commit info and add them to the pages. That way, when enabled each page will have the commit hash of its last modification displayed in the footer.

An example is available here: https://blog.hackervaillant.eu/about/

You can see that the footer references this [commit](https://github.com/gfeun/blog/commit//f14e076183fa134439d94c9d89200bfdd940c060)

If you go to https://blog.hackervaillant.eu/posts/test/ the footer points to a different [commit](https://github.com/gfeun/blog/commit//3aff2042d8751a3420ebfa0035913bfdb49d21e9) which introduced the post

To enable GitInfo you have to add the variable `enableGitInfo = true` in config.toml or pass --enableGitInfo at the command line

As before you also have to provide the commit base url, https://github.com/gfeun/blog/commit/ in my case
 
I also decided to remove the leading dot `·` because it renders as 
> ` Powered by Hugo & Coder.·[qsfsqf]` 

when one of credits or copyright is displayed. 

This is a breaking change because : 
- Passing environment variable is not necessary anymore 
- Displayed commit in the footer will be different for each page according to their last modification
- Main page will not have a commit hash since GitInfo only support content pages

### Issues Resolved

More consistent commit feature with the rest of the hugo ecosystem

### Checklist

#### General

#### Resources

#### Contributors

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -60,3 +60,4 @@
 - [Paolo Mainardi](https://paolomainardi.com)
 - [Ka-Wai Lin](https://github.com/kwlin)
 - [Piotr Orzechowski](https://orzechowski.tech)
+- [Glenn Feunteun](https://github.com/gfeun)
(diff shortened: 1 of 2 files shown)


luizdepra:
Looks good. Thank you.
I'll need to update the wiki about it, but later....

PR #298, opened 2020-04-08
Fix support for custom.css by adding tags

aristosvo:
### Prerequisites

### Description

It makes the custom_css working again

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -60,7 +60,8 @@
     {{ end }}
 
     {{ range .Site.Params.custom_css }}
-      <link rel="stylesheet" href="{{ . | relURL }}" />
+      {{ $style := resources.Get . }}
+      <link rel="stylesheet" href="{{ . | relURL }}" integrity="{{ $style.Data.Integrity }}" crossorigin="anonymous" media="screen" />
     {{ end }}
 
     {{ range .Site.Params.custom_js }}


aristosvo:
I’m running it now locally without any problems, but to be honest I haven’t done extensive testing!

aristosvo:
@clement-pannetier I wrapped it to only run when not local, does this work for you?

clement-pannetier:
> @clement-pannetier I wrapped it to only run when not local, does this work for you?

Yes, previously I had this error below :
image
What version of hugo do you have locally ?

layouts/_default/baseof.html
```diff
@@ -59,9 +59,15 @@
...
-    {{ range .Site.Params.custom_css }}
-      {{ $style := resources.Get . }}
-      <link rel="stylesheet" href="{{ . | relURL }}" integrity="{{ $style.Data.Integrity }}" crossorigin="anonymous" media="screen" />
+    {{ if .Site.IsServer }}
+      {{ range .Site.Params.custom_css }}
+        <link rel="stylesheet" href="{{ . | relURL }}"  media="screen" />
+      {{ end }}
+    {{ else }}
+      {{ range .Site.Params.custom_css }}
+        {{ $style := resources.Get . }}
```
clement-pannetier, line 68:
```suggestion
        {{ $style := resources.Get . | fingerprint}}
```

https://gohugo.io/hugo-pipes/fingerprint

aristosvo:
> What version of hugo do you have locally ?
Hugo Static Site Generator v0.68.3-157669A0/extended linux/amd64 BuildDate: 2020-03-24T12:13:38Z

aristosvo:
Probably due to you working on a local Windows environment. I've seen a few comments here and there mentioning it when looking for the solution myself.

clement-pannetier:
> Probably due to you working on a local Windows environment. I've seen a few comments here and there mentioning it when looking for the solution myself.

Maybe :)
Another thing, can you try changing `<link rel="stylesheet" href="{{ . | relURL }}"  media="screen" />`by `<link rel="stylesheet" href="{{ $style.RelPermalink }}" media="screen" />` for both conditions.
I have the same error as above, so can't test it :/

aristosvo:
It keeps working for me, both locally and remote. At your side?

clement-pannetier:
> It keeps working for me, both locally and remote. At your side?

No, same error as before, don't know why.
Be carefull, you deleted the `fingeprint` support.

aristosvo:
Does it work when you switch the changes back?

clement-pannetier:
> Does it work when you switch the changes back?

Yes, it's the $style.<whatever_variable> which causes problem.

aristosvo:
@clement-pannetier Did you configure custom_css in you config.toml? And where did you place the file for custom_css?

clement-pannetier:
> @clement-pannetier Did you configure custom_css in you config.toml? And where did you place the file for custom_css?

My config for custom css : `custom_css = ["css/custom.css"]`

aristosvo:
And do you have a file over there specified?

clement-pannetier:
> And do you have a file over there specified?

Yes of course !

aristosvo:
Okay, have you tried putting it in assets instead of static?

clement-pannetier:
> Okay, have you tried putting it in assets instead of static?

The resources/_gen/assets directory ?

clement-pannetier:
It's maybe the problem that you said earlier, it's a Windows problem...

aristosvo:
> The resources/_gen/assets directory ?

`{projectdir}/assets/css/custom.css`, so probably something like `clementpannetier.dev/assets/css/custom.css`

clement-pannetier:
> > The resources/_gen/assets directory ?
> 
> `{projectdir}/assets/css/custom.css`, so probably something like `clementpannetier.dev/assets/css/custom.css`

I don't have this folder, why do I have to create a new one only for css ?

aristosvo:
I'm sorry, I also struggled with it, should've mentioned it earlier. Here is an [explanation of Hugo itself](https://gohugo.io/getting-started/directory-structure/) and [here is a discussion](https://discourse.gohugo.io/t/the-base-path-for-resource-get/12823) about the basepath for `resource.Get`.

> assets
    Stores all the files which need be processed by Hugo Pipes. Only the files whose .Permalink or .RelPermalink are used will be published to the public directory. Note: assets directory is not created by default.

clement-pannetier:
> I'm sorry, I also struggled with it, should've mentioned it earlier. Here is an [explanation of Hugo itself](https://gohugo.io/getting-started/directory-structure/) and [here is a discussion](https://discourse.gohugo.io/t/the-base-path-for-resource-get/12823) about the basepath for `resource.Get`.
> 
> > assets
> > Stores all the files which need be processed by Hugo Pipes. Only the files whose .Permalink or .RelPermalink are used will be published to the public directory. Note: assets directory is not created by default.

Sorry for my incompetence, I'm relatively new to hugo.
As mentionned in the thread you provided me to, `assetDir = "static"`solved the issue.
I think I need to go deeper, but your links are a good start, thanks @aristosvo 👌

aristosvo:
Can you trigger the merge to master or do we have a dependency on an other contributor?

clement-pannetier:
> Can you trigger the merge to master or do we have a dependency on an other contributor?



> Can you trigger the merge to master or do we have a dependency on an other contributor?

@aristosvo Don't have the access right do do that.

clement-pannetier:
Hey @aristosvo, I went deeper on that. Why do you want to put the `custom.css` file inside `/assets` dir. The documentation says that this directory is meant to :
> Stores all the files which need be processed by Hugo Pipes.

We don't need Hugo Pipes for css files, it's usefull when you have SASS / SCSS files or images to process for example but that's not the case here !

Feel free to tell me if I'm wrong 😅

luizdepra:
I partially agree with @clement-pannetier here.

Adding integrity values for custom CSS files is a good thing, but Hugo's Asset Pipe doesn't look like the right way to do add this feature. Also, there is confusion about which folder should the files go.
I think we should add integrity values manually and optionally. We can try to support tuples or single values, something like this:

```
custom_css = ["css/custom.css"]
# or
custom _css = [("css/custom.css", "INTEGRITY_CODE")]
```

But I don't know if Hugo supports tuples. Needs checking.

BTW, I need to change these configuration names with `_` someday because Hugo uses camelCased names.

PR #317, opened 2020-04-26
Fix issue #314

### LINKED ISSUE #314: JavaScript references in wrong place
https://github.com/luizdepra/hugo-coder/blob/e46c452344a13d94b2748efa8a281778fc70df4b/layouts/_default/baseof.html#L66-L68

Scripts in head of html presentation should be at bottom of <body></body> element; 
> This prevents render blocking while the scripts load and is much better for site perception speed.
sources:
- https://stackoverflow.com/questions/3531314/should-i-write-script-in-the-body-or-the-head-of-the-html
- https://stackoverflow.com/questions/7751514/javascript-in-head-or-just-before-body

If anyone else agrees, I can make PR, although that'd take a little while. If you want to proceed with PR, go ahead.

clement-pannetier:
### Prerequisites

### Description

The `custom_js` script is now just before the end of the `</body>` element.

### Issues Resolved

#314

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -63,10 +63,6 @@
       <link rel="stylesheet" href="{{ . | relURL }}" />
     {{ end }}
 
-    {{ range .Site.Params.custom_js }}
-      <script src="{{ . | relURL }}"></script>
-    {{ end }}
-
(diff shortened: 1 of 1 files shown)


luizdepra:
Closes #314.

PR #335, opened 2020-05-17
Improve privacy by supporting gravatar MD5 hashes.

tkenvin:
### Prerequisites

### Description

Allows the site owner to use the pre-calculated MD5 hash instead of their email address for gravatar.

This improves privacy for the site owner if they want their site in a public repository but do not want their email address to be known.

The change is implemented by the addition of a new configuration parameter and a change to the logic when calculating the avatar image to use.

### Issues Resolved

This does not resolve any reported issue.

### Checklist

#### General

#### Resources

#### Contributors

### DIFF
--- a/layouts/partials/home.html
+++ b/layouts/partials/home.html
@@ -1,12 +1,17 @@
 <section class="container centered">
   <div class="about">
-    {{ if and (isset .Site.Params "avatarurl") (not (isset .Site.Params "gravatar")) }}
+    {{ if (isset .Site.Params "gravatar") }}
+      {{ with .Site.Params.gravatar }}
+        <div class="avatar"><img src="https://www.gravatar.com/avatar/{{md5 .}}?s=240&d=mp" alt="gravatar"></div>
+      {{ end }}
+    {{ else if (isset .Site.Params "avatarurl") }}
       {{ with .Site.Params.avatarurl }}
         <div class="avatar"><img src="{{ . | relURL }}" alt="avatar"></div>
       {{ end }}
-    {{ end }}
-    {{ with .Site.Params.gravatar }}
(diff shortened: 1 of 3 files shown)


layouts/partials/home.html
```diff
@@ -1,12 +1,17 @@
...
+    {{ else if (isset .Site.Params "avatarurl") }}
       {{ with .Site.Params.avatarurl }}
         <div class="avatar"><img src="{{ . | relURL }}" alt="avatar"></div>
       {{ end }}
-    {{ end }}
-    {{ with .Site.Params.gravatar }}
-      <div class="avatar"><img src="https://www.gravatar.com/avatar/{{md5 .}}?s=240&d=mp" alt="gravatar"></div>
+    {{ else if (isset .Site.Params "gravatarmd5") }}
+      {{ with .Site.Params.gravatarmd5 }}
+        <div class="avatar"><img src="https://www.gravatar.com/avatar/{{ . }}?s=240&d=mp" alt="gravatar"></div>
```
luizdepra, line 13: Both URLs, for `gravatar` and `gravatarmd5`, are the same, right?
It looks like if I put an MD5 hash in the `gravatar` config it will work too.
So, there is no need to add another configuration here, we just need to document this behavior.

tkenvin: I'm new at this so there may be a better way of doing it but there is a distinction between `gravatar` and `gravatarmd5` URLs:
`gravatar` runs `{{ md5 . }}` whereas `gravatarmd5` simply uses Dot `{{ . }}`.

luizdepra: Ok, I got it wrong.
There is better wait to handle "secrets", provided by Hugo, without changing the theme You can use an environment variable. In this case, it would be `HUGO_PARAMS_GRAVATAR`.

Issue #237, opened 2019-11-25
i18n translation used in layout files


tobiasehlert:
As right now, there are some translations in for example [i18n/de.toml](https://github.com/luizdepra/hugo-coder/blob/master/i18n/de.toml)

Should not files like [layouts/posts/list.html](https://github.com/luizdepra/hugo-coder/blob/master/layouts/posts/list.html) use the tags then, or maybe use newly added ones?

Would be awesome if the "Posts" word would be localized in the i18n file, if that now is the correct way to go. What else would the [i18n/de.toml](https://github.com/luizdepra/hugo-coder/blob/master/i18n/de.toml) file be useful for?

Maybe a little related to [issue 70](https://github.com/luizdepra/hugo-coder/issues/70)..

luizdepra:
Yes, I probably need a `i18n` tag in the posts list template. And, of course, we need to add a translation entry in every language file.
I'm not 100% sure if this is everything we need to do, but it seems so. I need to test first.

Issue #70, opened 2018-08-03
How to change /posts page title?


vitalyavolyn:
Hi! How can i change Title "Posts" on /posts page? I want to translate it into my native language.

Also, can i make dates on same page look like DD.MM.YYYY? Thanks.

vitalyavolyn:
Just found out that i can override partials by copying them in layouts/partials/ 
I guess i need to do it this way, right?

luizdepra:
Yep. It should work this way.
You will need to do same with the date, I think. I don't remember if it is possible to override the date format with configuration.

vitalyavolyn:
I was able to change date format, thanks. But heading? 

layouts/partials/list.html
```html
<h1 class="title">{{ .Title }}</h1>
```

Where does `.Title` come from?

luizdepra:
It comes from the archetype `posts`.

To accomplish that probably you need to put all your post inside a folder with the desired name.
Maybe, you will need make a copy of the post related files with the desired name too.

vitalyavolyn:
Thanks for your help!

kshitij10496:
An alternate way to just capitalize the heading from "Posts" to "POSTS" would be to create a `_index.md` file within the `content/posts` directory using `hugo new posts/_index.md`.
We can now manually edit the `title` page variable as we like.

PR #337, opened 2020-05-18
Allow internationalization for 'Posts'

latiif:
### Prerequisites

### Description

Allows for customization in the internationalization files for "Posts". 

If there is an entry in the `i18n/` files under 'Posts' (Uppercase P), it will be used in the `posts` page. 
If not, it will revert to what would've been displayed instead, and not create a breaking change.

**NOTE** While this is NOT a breaking change, we still need to update the `i18n/` files for the changes to reflect.

### Issues Resolved

How to change /posts page title? #70 
i18n translation used in layout files #237

### DIFF
--- a/layouts/posts/list.html
+++ b/layouts/posts/list.html
@@ -3,7 +3,13 @@
 {{ end }}
 {{ define "content" }}
   <section class="container list">
-    <h1 class="title">{{ .Title }}</h1>
+    <h1 class="title">
+      {{- if i18n .Title}}
+        {{- i18n .Title }}
+      {{ else }}
+        {{- .Title }}
+      {{ end }}
+    </h1>
 
     <ul>
       {{- range .Paginator.Pages -}}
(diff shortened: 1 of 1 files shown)


layouts/posts/list.html
```diff
@@ -3,7 +3,13 @@
 {{ end }}
 {{ define "content" }}
   <section class="container list">
-    <h1 class="title">{{ .Title }}</h1>
+    <h1 class="title">
+      {{- if i18n .Title}}
```
luizdepra, line 7: I think is better to do something like this `{{ title (i18n (lower .Title)) | default .Title }}`. So, you should put "posts" as id in the language file. Yep, lower case.
Also, you need to translate the title inside the other `define`. This is the title shown in the browser tab/topbar.

luizdepra:
Nothing that I can remember, @latiif. But there is no problem with little repetitions in this case.

a1x42:
@latiif other translations are capitalised when displayed e.g. my translation file contains téma for tag but on the list page for a specific tag it says Téma. For the posts translation this does not happen. could this be made consistent please?

latiif:
You are correct. This was an oversight as I needed/tested this for Arabic which does not have lowercase/uppercase. A fix is suggested in #338.

Issue #350, opened 2020-06-03
how to add light and dark mode switching button in the index page


iamrajiv:
how to add light and dark mode switching button in the index page
<img width="88" alt="Screenshot 2020-06-03 at 5 54 57 PM" src="https://user-images.githubusercontent.com/42106787/83636304-59adf580-a5c3-11ea-944f-ab2f7205f3e2.png">
I have edited `config.toml` by adding `colorscheme = "auto"` but I am not getting it.

clement-pannetier:
Hi @rajmahendra, this feature has been discussed here #226 but the "automagic" solution has been chosen. We chose the prefers-color-scheme approach.

You can see an implementation of this functionnality on my GitLab repo: https://gitlab.com/clement-pannetier/clementpannetier.dev and see it directly on my website: https://clementpannetier.dev

**Just a quick note, you're maybe going to experience FOUC problems when switching themes**

nturaga:
Is there a PR for this?

clement-pannetier:
> Is there a PR for this?

No

nturaga:
@clement-pannetier Can you highlight which parts of the code are the ones needed specifically? I'm unable to get it working on mine.

clement-pannetier:
@nturaga ⬆😉

Issue #327, opened 2020-05-10
Switching between dark and light background


rajmahendra:
Provide a button for dark and light background. when the viewer clicks a toggle the site changes its background to dark or light respectively

clement-pannetier:
Hi @rajmahendra, this feature has been discussed here #226 but the "automagic" solution has been chosen.

luizdepra:
BTW, this is kinda easy to customize.
But we go with the `prefers-color-scheme` approach.

luizdepra:
Well, We've merged this feature while ago.

PR #358, opened 2020-06-28
Add theme toggle button feature

clement-pannetier:
### Prerequisites

### Description

I added this feature because a lot of peoples asked me how I did it on my website and some, despite the link to my gitlab repo, didn't get it working.
Issues #350 #327 for example.

#### Resources


exampleSite/config.toml
```diff
@@ -53,7 +53,7 @@ disqusShortname = "yourdiscussshortname"
     custom_css = []
 
     # Custom JS
-    custom_js = []
+    custom_js = ["js/custom.js"]
```
luizdepra, line 56: Is better to add this JS file into the <head> scope. Let's leave the custom_js field to user's scripts.

layouts/_default/baseof.html
```diff
@@ -90,13 +90,7 @@
...
   </head>
 
-  {{ $csClass := "colorscheme-light" }}
-  {{ if eq .Site.Params.colorscheme "dark" }}
-    {{ $csClass = "colorscheme-dark" }}
-  {{ else if eq .Site.Params.colorscheme "auto" }}
-    {{ $csClass = "colorscheme-auto" }}
-  {{ end }}
-  <body class="{{ $csClass }}{{ if .Site.Params.rtl }} rtl{{ end }}"
+  <body class="{{ if .Site.Params.rtl }} rtl{{ end }}"
```
luizdepra, line 93: I think we still need these color scheme configurations. I want to set a default color theme but let the user change it.
We'll need to run a JS function on load to change the color scheme if the user preference is stored in Local Storage.

layouts/partials/header.html
```diff
@@ -4,6 +4,13 @@
       {{ .Site.Title }}
     </a>
     {{ if or .Site.Menus.main .Site.IsMultiLingual }}
+    <span id="dark-mode-toggle" class="menu-button float-right">
+      <svg width="24" height="24" viewBox="0 0 48 48" fill="none" xmlns="http://www.w3.org/2000/svg">
```
luizdepra, line 8: We can use FontAwesome icons instead of a custom SVG.
https://fontawesome.com/icons/adjust?style=solid

exampleSite/config.toml
luizdepra: Like CSS files, we should minify this JS file too.

luizdepra:
The code seems ok, but the toggle button is not working in the PR deploy.
https://deploy-preview-358--hugo-coder.netlify.app/

I'm going to take a look later and try to find the problem, ok?

clement-pannetier:
Hey @luizdepra, I found the problem, it's related to the `colorsheme-auto` class.
The toggle button works until you go to another page or reload it. It's because I only remove this class at first launch (line 12). First launch = no localstorage item set.

Could you explain how the toggle button should behave depending on the value of the configuration variable ? Thanks.

luizdepra:
No problem. Let's see. I think we should respect users' choices and let them always change the color mode. So, when loading a page:
- If there is a preference stored in local storage, load it.
- If not and the user has `prefers-color-scheme` defined, load it and save in local storage.
- If not fallback to default color scheme (set in `config.toml`) and save it in local storage.

When clicking in the color mode toggle we should change the color mode and save it in local storage.

clement-pannetier:
@luizdepra, I fixed this the simple way 😅, and also did a simple refactor.

> If not and the user has `prefers-color-scheme` defined, load it, and save in local storage.

The `prefers-color-sheme` cannot be not defined, it's either `dark` or `light`.

exampleSite/config.toml
```diff
@@ -53,7 +53,7 @@ disqusShortname = "yourdiscussshortname"
     custom_css = []
 
     # Custom JS
-    custom_js = []
+    custom_js = [""]
```
luizdepra, line 56: Just one thing now.
```suggestion
    custom_js = []
```
Leaving an empty string here causes this:
```
<script src="https://deploy-preview-358--hugo-coder.netlify.app/"></script>
```
An `script` entry for a non-existing JS file point to the base URL.

luizdepra: Well, I can commit to your branch, so I just did it.

luizdepra:
Thank you again and sorry for taking to long to review.
(The pandemic situation, too much work and family concern are degrading my time and will to work in side projects. :disappointed: )

clement-pannetier:
> Thank you again and sorry for taking to long to review.
> (The pandemic situation, too much work and family concern are degrading my time and will to work in side projects. 😞 )

No problem, take care 😉

PR #370, opened 2020-07-17
Adds romanian translation

sdwolfz:
### Prerequisites

### Description

Adds Romanian translation. Tried it with my website to see how the reading time behaves, looks to be working well.

### DIFF
--- a/i18n/ro.toml
+++ b/i18n/ro.toml
@@ -1,0 +1,24 @@
+[category]
+other = "categorie"
+
+[tag]
+other = "etichetă"
+
+[series]
+other = "serie"
+
+[reading_time]
+one   = "Un minut de lectură"
+other = "{{ .Count }} {{ if lt .Count 20 }}minute{{ else }}de minute{{ end }} de lectură"
+
+[page_not_found]
+other = "Pagină Pierdută"
+
+[page_does_not_exist]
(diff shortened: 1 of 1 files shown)


i18n/ro.toml
```diff
@@ -0,0 +1,23 @@
...
+other = "categorie"
+
+[tag]
+other = "etichetă"
+
+[series]
+other = "serie"
+
+[reading_time]
+other = "{{ .Count }} {{ if eq .Count 1 }}minut{{ else if lt .Count 20 }}minute{{ else }}de minute{{ end }} de lectură"
```
luizdepra, line 11: Have you tried to use `one`, `few`, `many`, and `other` translations here?
It works fine to Slovak. Maybe Romanian is the same case.

https://github.com/luizdepra/hugo-coder/blob/master/i18n/sk.toml#L10-L14

sdwolfz: I'll need to look into the cut-over for `few` and `many`. It will work for `one`, but the special form is for elements between `[2,19]`.

sdwolfz:
@luizdepra I've pushed a change to use `one`, and `other`. Can't find how `few` and `many` are defined.

luizdepra:
@sdwolfz, no problem.
Thank you.

PR #375, opened 2020-07-18
Separates page title from navtitle

sdwolfz:
### Prerequisites

### Description
This is how it looks on my website:

Screenshot_2020-07-18_14-11-15

The reason for this change is that my name is long, and having it appear twice on the main page is too much, so I put my online alias there instead, and removed the forced uppercase, so it can be customised by users just the way they like it. It also fits in with the website domain. Also, for the title I have both my name and alias, but the name is without diacritics, since that helps with SEO.

Was thinking I could get this upstreamed.

> Also note, I've changed the `info` to be an `<i>` instead of a `<h2>` since I like that look better (there's too much bold on the page). @luizdepra do you think I should open a PR with that change too?


assets/scss/_navigation.scss
```diff
@@ -17,7 +17,6 @@
   }
   .navigation-title {
     letter-spacing: 0.1rem;
-    text-transform: uppercase;
```
luizdepra, line 20: Revert this, please.
Let's keep the original style.

layouts/partials/header.html
```diff
@@ -1,7 +1,7 @@
 <nav class="navigation">
   <section class="container">
     <a class="navigation-title" href="{{ .Site.BaseURL | relLangURL }}">
-      {{ .Site.Title }}
+      {{ .Site.Params.Navtitle }}
```
luizdepra, line 4: We should have a fallback here.

```
{{ .Site.Params.Navtitle | default .Site.Title }}
```

sdwolfz:
@luizdepra actually I've decided I'll keep this change on my fork only.

Issue #361, opened 2020-07-07
Switch to Fork Awesome, or make the font location customizable


paride:
While free to use, Font Awesome is not free software. Fork Awesome (https://forkaweso.me/) can mostly replace it, and it's MIT licensed like hugo-coder.

Bonus point: it has a Debian icon, which Fork Awesome lacks :)

If you are not willing to switch, then it would be nice to have the possibility to override the font css location set in `baseof.html` with a custom location from the `config.toml` file.

luizdepra:
No objection. Would be nice if you create a PR to change it.

sdwolfz:
On my fork I've removed the hardcoded Font Awesome and am serving them from assets, was thinking of implementing this but did not know about Fork Awesome.

Give me a weekend and I might be able to open a PR.

metbril:
I second the option to locally host my fonts (no tracking, no external dependencies).

luizdepra:
We have a branch with locally hosted fonts, scripts, etc. It can be used as a base for these changes, but it is really outdated.

luizdepra:
Doing to put this into v2 roadmap.

luizdepra:
Let's continue the discussion at #412.

PR #376, opened 2020-07-18
Local fonts and css support

sdwolfz:
### Prerequisites

### Description

Closes: #361

Instead of using remote fonts and css/scss, have it customizable.

### DIFF
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -49,8 +49,17 @@ disqusShortname = "yourdiscussshortname"
     # Enable Twemoji
     enableTwemoji = true
 
+    # Remote font and css
+    use_default_remote_font = true
+    use_default_remote_css  = true
+
     # Custom CSS
-    custom_css = []
+    custom_css_pre  = []
+    custom_css_post = []
+
+    # Custom SCSS
+    custom_scss_pre  = []
+    custom_scss_post = []
 
     # Custom JS
     custom_js = []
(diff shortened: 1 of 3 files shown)


exampleSite/config.toml
```diff
@@ -49,8 +49,17 @@ disqusShortname = "yourdiscussshortname"
     # Enable Twemoji
     enableTwemoji = true
 
+    # Remote font and css
+    use_default_remote_font = true
```
metbril, line 53: In general, Hugo parameters do not have underscores in their names (like `enableTwemoji` instead of `enable_twemoji`. There isn't a style guide, but it is common use.

layouts/_default/baseof.html
```diff
@@ -26,9 +26,30 @@
       <link rel="canonical" href="{{ .Permalink }}">
     {{ end }}
 
-    <link href="https://fonts.googleapis.com/css?family=Lato:400,700%7CMerriweather:300,700%7CSource+Code+Pro:400,700&display=swap" rel="stylesheet">
```
metbril, line 29: I'm not a CSS guru. Why can't the change just exclude remote assets by a parameter? Then the already included custom CSS option can embed the local fonts, CSS etc.?

sdwolfz: There is `use_default_remote_css` below. I'll change it to exclude only when a user wants custom ones.

luizdepra:
We are adding too much complexity to this theme configuration. I prefer to change the theme to always use self-hosted fonts and other dependencies. 

We already started this movement in the [self-hosting-deps branch](https://github.com/luizdepra/hugo-coder/tree/self-hosting-deps). I just need to come back there and finish it.

sdwolfz:
@luizdepra OK I'll keep these changes on my fork.

PR #377, opened 2020-07-18
Adds customizable favicons

sdwolfz:
### Prerequisites

### Description

Makes favicons customizable, since there are many possibilities for them: https://stackoverflow.com/a/26768184

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -80,8 +80,9 @@
       {{ end }}
     {{ end }}
 
-    <link rel="icon" type="image/png" href="{{ .Site.Params.favicon_32 | default "/images/favicon-32x32.png" | absURL }}" sizes="32x32">
-    <link rel="icon" type="image/png" href="{{ .Site.Params.favicon_16 | default "/images/favicon-16x16.png" | absURL }}" sizes="16x16">
+    {{ range .Site.Params.favicons }}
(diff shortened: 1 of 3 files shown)


layouts/_default/baseof.html
```diff
@@ -80,8 +80,9 @@
       {{ end }}
     {{ end }}
 
-    <link rel="icon" type="image/png" href="{{ .Site.Params.favicon_32 | default "/images/favicon-32x32.png" | absURL }}" sizes="32x32">
-    <link rel="icon" type="image/png" href="{{ .Site.Params.favicon_16 | default "/images/favicon-16x16.png" | absURL }}" sizes="16x16">
+    {{ range .Site.Params.favicons }}
+      <link rel="{{ .rel }}" type="{{ .type }}" href="{{ .href | absURL }}" sizes="{{ .sizes }}">
```
metbril, line 84: This change causes that favicons is a required config item? In my opinion the template should provide sensible defaults that can be overwritten with an optional parameter.

metbril: It also introduces a **breaking change**, since the `favicons` is a new config parameter.

sdwolfz: How about default the initial two, else if custom settings are specified, my changes?

luizdepra:
Hey, this is cool but I think this theme already has too many configurations options that could be made simply by change a file or two.
You change the favicons by replacing the default ones. Also, you could add more icons by modifying `baseof.html`.

sdwolfz:
@luizdepra OK I'll keep these changes on my fork.

PR #382, opened 2020-08-02
Add support for plausible analytics

rhysperry111:
This PR should add support for [plausible](https://plausible.io) analytics.

It is just a modified copy of the PR that implemented fathom analytics


luizdepra:
> Hi @rhysperry111, there is already google and fathom analytics for this theme. I think @luizdepra is gonna say that "this is cool but I think this theme already has too many configurations options that could be made simply by change a file or two." .
> Also, when adding new config parameters, you should also add them to the `stackbit.yaml` file.

This is not the case, @clement-pannetier. hahaha
I try to avoid adding styling and other simple customizations to this theme, but adding a new analytics tool support is kinda fine.

luizdepra:
Thank you, @rhysperry111.

PR #394, opened 2020-08-17
Switched to jsdelivr for css and js

rabinadk1:
### Prerequisites

- [ ] This pull request introduces a breaking change.

### Description

Jsdelivr is a better choice because it uses multiple CDNs and DNS servers. I also added version aliasing which enables users to receive bug fixes and new non-breaking features.

### Issues Resolved

### Checklist

#### General

#### Resources

#### Contributors

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -73,3 +73,4 @@
 - [Pakhomov Alexander](https://github.com/PakhomovAlexander)
 - [Rhys Perry](https://rhysperry.com)
 - [Arunvel Sriram](https://github.com/arunvelsriram)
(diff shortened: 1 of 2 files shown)


luizdepra:
We should apply [SRI](https://www.jsdelivr.com/using-sri-with-dynamic-files) for all these files included.

Also, I need to mention that I will migrate all dependencies to be self-hosted in a near future. So, these changes will be lost.

rabinadk1:
Using SRI stops us from getting minor updates from the server which are good. I think we can trust jsdelivr on that.

luizdepra:
Thank you, but I prefer to keep dependencies with fixed version and SRI.

Issue #385, opened 2020-08-12
Long post titles wrap under the date in the list view


jsturdy:
I noticed that if there is a post with an exceptionally long title, in the list page, when it wraps, it flows under the date column.
To me this was visually off, as I'd want the title and date to be separate columns
Example 
![original-with-long-title](https://user-images.githubusercontent.com/5197820/90027702-1e156300-dcb9-11ea-8892-5d8cca9e1e9f.png)

In the mobile/reduced viewport, this collapses well (by design I expect):
![original-mobile-with-long-title](https://user-images.githubusercontent.com/5197820/90027799-3ab19b00-dcb9-11ea-8ac9-d36cadc33eac.png)

What I was thinking would be something like this (wide viewport):
![modified-with-long-title](https://user-images.githubusercontent.com/5197820/90027878-5026c500-dcb9-11ea-836c-3fca3eb7268a.png)

Then on a mobile/small viewport there are two paths one could take:
1) The default would look no different from the current mobile view, making the transition at a screen width of 768
1) However at that size, one can still appreciate the split columns as seen here (this still includes the increased spacing and left-justification of the date, that may or may not be desired if the columns are still separated, but it worked visually for me, so I didn't change it):
![modified-standard-mobile-with-long-title](https://user-images.githubusercontent.com/5197820/90028086-8cf2bc00-dcb9-11ea-9961-92dea96ecd4d.png)

and one could make a change to transition to the collapsed view at a narrower screen (not sure what would be "ideal" but something around 500 worked well enough in my quick tests
![modified-small-mobile-with-long-title](https://user-images.githubusercontent.com/5197820/90028242-b7dd1000-dcb9-11ea-970b-2542b206cc14.png)

If any of this is interesting, I'd be happy to submit a PR.
The baseline change of fixing the flow is a simple three-liner, while the additional handling is only minimally more change, but might need some inputs to decide how best to present it.

luizdepra:
Hey!
A PR to fix this issue would be nice, but I prefer to keep the mobile style as it is now (date above and title below).

ketan-vijayvargiya:
This commit broke the pagination links on desktop. (Mobile is fine.)

Scroll down to the bottom for a live example: https://ketanvijayvargiya.com/notes/

[...]

PR #407, opened 2020-09-19
[bugfix] Fix problem with pagination from wrapped post titles fix

jsturdy:
### Prerequisites

### Description

In the followup to #385 it is noted that the fix I submitted in #401 introduced a bug in the pagination layout.
This PR fixes that issue by moving the line-wrapping media query into a block that is applied *only* if the element isn't inside the `.pagination` class.

### Issues Resolved

Resolves followup issue from merging of #401 as noted in #385

### Checklist

#### General

#### Resources

#### Contributors


luizdepra:
Nice. Thx.

PR #411, opened 2020-09-25
Use relative URLs everywhere and remove <base> tag

autarch:
### Prerequisites

### Description

There were a few places still using absolute URLs, including `<link rel="icon">` tags, `<link rel="alternate">` tags, and nav menu links. In addition, the presence of a `<base>` tag caused the browser to absolute-ize all relative URLs.

This is a problem for things like render.com, which will put the site on its own domain for previews. For example, I have a preview site at https://houseabsolute-com.onrender.com/. Ultimately this will be served at https://houseabsolute.com/, but before I switched over I wanted to preview the site.

The absolute URLs were being generated for https://houseabsolute.com/ because of my baseURL setting in config and the presence of the <base> tag. This completely breaks the preview site.

### Issues Resolved

None that I could find.

### Checklist

#### General

#### Resources

#### Contributors

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -76,3 +76,4 @@
 - [Lorenzo Cameroni](https://github.com/came88)
 - [Jared Sturdy](https://github.com/jsturdy)
 - [Daniel Monteiro](https://github.com/dfamonteiro)
+- [Dave Rolsky](https://github.com/autarch)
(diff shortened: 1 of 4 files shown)


luizdepra:
Hmm. Looks good.
But I think we should remove `canonifyURLs = true` from the example config too.

autarch:
> But I think we should remove `canonifyURLs = true` from the example config too.

Good catch. I force pushed an update that includes this.

luizdepra:
Nice. Thank you.

Issue #418, opened 2020-09-28
[v2] Floating Color Scheme Toggle


luizdepra:
Move the toggle icon to be always floating in the bottom-left corner of the screen. Something like [this site](https://fasterthanli.me/articles/so-you-want-to-live-reload-rust) has.

JianLoong:
Sorry for the spam. Didn't notice there were other things to be done until I took a look at the PR template. The Netlify preview seems fine. The only thing that bothers me is the size of the toggle button itself.

PR #433, opened 2020-10-09
Moved color scheme toggle

JianLoong:
### Prerequisites

- [x]  This pull request fixes a bug.

### Description

Describe what this pull request achieves.

Moved the position of the color scheme toggle to the bottom left of the screen.

### Issues Resolved

hugo-coder-normal

hugo-coder-responsive

### Checklist

#### General

- [x] Reference issue with #418 if applicable

#### Resources

#### Contributors


assets/scss/_base.scss
```diff
@@ -217,3 +217,13 @@ img.emoji {
...
 }
+
+#dark-mode-toggle {
+  bottom:0;
+  left:0;
+  z-index: 100;
+  position: fixed;
+  width:3em;
+  height:3em;
+  font-size: 1.6em;
```
luizdepra, line 228: You can remove both `width` and `height` rules here. And set `bottom` and `left` to `2rem`.

layouts/partials/footer.html
```diff
@@ -22,5 +22,10 @@
         {{ end }}
       {{ end }}
     </section>
+    {{ if not .Site.Params.hidecolorschemetoggle }}
+    <span id="dark-mode-toggle" class="">
```
luizdepra, line 26: I'm missing the on-hover effect here. Maybe changing from `span` to `a` should be enough. Also, please remove these empty `class` attribute.

luizdepra:
I think we should "steal" [LoveIt](https://themes.gohugo.io//theme/LoveIt/theme-documentation-basics/)'s Style for these buttons and put a rounded background under the icon.

Tell me if you have time to make these changes. If not, I can merge your changes in a separated branch and do all the fixes.

JianLoong:
I fixed this issue.

However, previously because the colour scheme toggle was on the navigation bar, thus SCSS is under the navigation category. So for now, I just changed it there. 

Visually however, the toggle seems to be under footer section partial but, the HTML for it is **still** under the **header.html** partial. 

Do you want it to moved into the **footer.html** partial instead? Which approach is more logical? Seems to me, that the question "Is colour toggle considered a navigation element?" arises.

For now, I opted to leave the footer partial as it is and made minor changes to the head.html and changed SCSS related to the color toggle in the navigation SCSS.

Feel free to merge the branch and obtain "ideas" from the other style;  :smirk:

luizdepra:
I think the toggle button should be neither in the header nor in the footer. It should be in a proper div inside the body. But, don't worry about that, I can move it later.

I'm going to merge this PR into a new branch this weekend and work on it. Thank you in advance. :)

Issue #416, opened 2020-09-28
[v2] Information Panel Shortcodes


luizdepra:
Add Hugo's shortcodes for information panels: info, tip, note, warning and error.
See [Jira's Note Macros](https://confluence.atlassian.com/doc/info-tip-note-and-warning-macros-51872369.html) as model.

luizdepra:
Done at #537.

PR #437, opened 2020-10-15
Added information panel shortcodes

JianLoong:
### Prerequisites

### Description

Describe what this pull request achieves.

I actually took a stab at #416 because I was interested to see how the information panels and alert would look on a dark theme.

I added a post on it so that it possible to take a look at them at the deployment preview. At the moment, it seems to be working out OK, but the colours are still quite iffy. The main ideas were taken from [here](https://github.com/martignoni/hugo-notice)

However, the SVG elements were removed and replaced by fork awesome icons. 

### Issues Resolved

### Checklist

#### General

- [x] Reference issue with #416  if applicable

#### Resources

#### Contributors

Edit : I know it doesn't look exactly like the example, but I just wanted to take a stab at it.


luizdepra:
Nice. This is very good.
But I think we can change the style a bit. I'll come back here later this weekend with some ideas.

JianLoong:
Sure. There was an issue with the light theme colour in the previous push as I was messing around with the dark theme.

The other thing is that, should the usage of markdown be allowed in the info panel itself? 

One thing I have noticed is that if that is enabled, something like this will happen..

```
{{< notice note >}}
    This does not start at the beginning of as there is tab in front.
{{< /notice >}}
```

Essentially a ``<pre><code>`` will be rendered at the HTML, not sure why this would happen. 

Also there is also support for il8n, for users to configure the "note", "warning", "tip" and "info" messages.

By the way, I really like this theme its really clean both in the looks and the codes as well. :clap:

luizdepra:
Well, I'm going to merge this PR into a separated branch and work on it. Thank you, @JianLoong.

PR #439, opened 2020-10-15
Adds anchors to all headings

sdwolfz:
### Prerequisites

### Description

Adding anchor links to all headings. ~Hovering over a heading (h1..h∞) will display a clickable `#` which will change the URL to contain a reference to the anchor~. Headings are now links that change the URL to contain a reference to the anchor. Great for sharing links to specific content on a page. I've added it everywhere except the home page h1.

Support was added in this PR: https://github.com/gohugoio/hugo/pull/7133/files

You can see this working on my website (link in my profile).

It's a bit rough around the edges, please advise on how to improve the CSS.


sdwolfz:
Update: `#` links on the left side looked awkward, and displayed outside of the screen on mobile, so I rewrote the functionality to instead make all headings have links with anchors.

luizdepra:
I think we could have a different approach. I'm thinking if we use on hover to display a link icon on the left side of the title and only this icon has the link to the anchor. Something like how [Terminal Theme](https://themes.gohugo.io//theme/hugo-theme-terminal/post/markdown-syntax/) does.

Also, we can leave the page's main title without an anchor.

sdwolfz:
@luizdepra you mean on the right side? Yes that might work better, I'll try it out and see how it looks.

I also see Terminal has the main h1 link to the page's URL. I kind of like that as a functionality, do you think you would be OK to keep that?

sdwolfz:
@luizdepra can you please take another look at this?

luizdepra:
Sorry, for taking to long to come back here. I'm recovering from a burnout, so I'm taking easy and coming back slowly.

So...

> you mean on the right side? Yes that might work better, I'll try it out and see how it looks.

Yup, you nail it.

> I also see Terminal has the title headings link to the page's URL. I kind of like that as a functionality, do you think you would be OK to keep that?

It could look nice too. I just would like to change the hash icon to a [link icon](https://forkaweso.me/Fork-Awesome/icon/link/). What do you think?

layouts/_default/_markup/render-heading.html
```diff
@@ -0,0 +1,4 @@
+<h{{ .Level }} id="{{ .Anchor | safeURL }}">
+  {{ .Text | safeHTML }}
+  <a class="heading-link" href="#{{ .Anchor | safeURL }}">⌗</a>
```
luizdepra, line 3: I think we should use a font awesome icon here instead.

sdwolfz:
No worries, take as much time as you need.

I'll use the link icon :+1: Will push on my fork first and see how that looks, then I'll update this PR.

sdwolfz:
All done @luizdepra, you can see it in action here: https://www.codrut.pro/snippets/how-to-use-meson/

Not sure what Netlify's problem is in CI...

luizdepra:
> Not sure what Netlify's problem is in CI...

Yup. It was some cache problem. Just refreshed it and it is fine now.

About the PR, everything is fine. Thank you. :)

PR #333, opened 2020-05-15
Minor CSS bug fix

ajdintrejic:
### Prerequisites

Put an `x` into the box(es) that apply:

- [x] This pull request fixes a bug.
- [ ] This pull request adds a feature.
- [ ] This pull request introduces breaking change.

### Description

This pull request fixes a bug that happens on mobile devices when you open and close the hamburger menu and it makes the hamburger menu blue.

image

I have removed the CSS selectors which were causing the bug. I have no idea why they were there.

### Issues Resolved
https://github.com/luizdepra/hugo-coder/issues/329

### Checklist

Put an `x` into the box(es) that apply:

#### General

- [ ] Describe what changes are being made
- [ ] Explain why and how the changes were necessary and implemented respectively
- [ ] Reference issue with `#<ISSUE_NO>` if applicable

#### Resources

- [x] If you have changed any SCSS code, run `make release` to regenerate all CSS files

#### Contributors

- [ ] Add yourself to `CONTRIBUTORS.md` if you aren't on it already

I'm not on the list, I'm not sure if I did contribute enough to be there.

PR #446, opened 2020-10-19
Fix burger focus color change

sdwolfz:
### Prerequisites

### Description

Reference: https://github.com/luizdepra/hugo-coder/pull/333

:point_up: Fixing the bug detailed in the PR above by forcing the css of the burger menu button to not have a focus color change.


luizdepra:
Nice. Could you please update the generated css files with `make release` and add these new files to this PR?

sdwolfz:
Running `make release` simply deletes files in `resources/_gen`. Is this the intended result? (if yes I can squash).

Also, please don't put a hacktoberfest label on this PR, I don't participate in that spam.

sdwolfz:
:point_up: Never mind, I was compiling with arch hugo instead of the extended version. Pushed the changes. Let me know if it's OK.

Also, do you want a docker image for the project? Here's the one I use: https://gitlab.com/sdwolfz/docker-images/-/blob/master/hugo/Dockerfile

You can run it with `docker run --rm -it sdwolfz/hugo:latest sh` but I think it's better to have it part of the project, and it does not have rsync as I don't use that.

luizdepra:
Well, this project is just a theme. Adding docker to build its demo is kinda overkill and one more thing to update (or forget to). So, I prefer to not include a dockerfile here. 

About your PR. I tried to test in on my phone but the problem is still there. Maybe my phone is caching something. I'll try it again later.

luizdepra:
I think I found the problem. We need to add the same rule into `_navigation_dark.scss`.

sdwolfz:
Firefox touch simulation was not triggering the rule so I thought it worked :see_no_evil: 

Pushed dark mode changes.

luizdepra (approved):
Nice. Now it is working fine.
Thank you.

Issue #417, opened 2020-09-28
[v2] Fix missing i18n strings


luizdepra:
There are some missing translations. We should find them, translate and add default values.

clement-pannetier:
Recap of missing translations:
- [ ] **Arabic**: author, see_also, posts @latiif 
- [ ] **Bangali**: author @arafat-hasan 
- [ ] **Czech**: author, posts @tomkys144 
- [ ] **German**: author, see_also, posts @AltNico 
- [ ] **Spanish**: author, see_also, posts @viticlick @AltNico 
- [ ] **Finnish**: author, see_alo, posts @endormi 
- [ ] **French**: posts @clement-pannetier 
- [ ] **Hebrew**: author, see_also, posts @FIGBERT 
- [ ] **Hindi**: author, see_also, posts @iamrajiv 
- [ ] **Hugarian**: author, see_also, posts @voroskoi 
- [ ] **Italian**: author, posts @came88 
- [ ] **Japanese**: author, see_also, posts @josephting 
- [ ] **Malay**: author, posts @hussaini 
- [ ] **Dutch**: author, see_also, posts @metbril 
- [ ] **Polish**: author, see_also, posts @0rzech 
- [ ] **Brazilian Portuguese**: author, see_also, posts @luizdepra 
- [ ] **Romanian**: author, see_also, posts @sdwolfz 
- [ ] **Russian**: author, see_also, posts @PakhomovAlexander @estim 
- [ ] **Swedish**: author, see_also, posts @tobiasehlert 
- [ ] **Slovak**: author, see_also, posts @a1x42 
- [ ] **Turkish**: author, see_also, posts @tosbaha 
- [ ] **Simplified Chinese**: author, see_also, posts @PanZiwei 
- [ ] **Taiwan Chinese**: author, see_also, posts @TMineCola 

Help would be appreciated 👍.

a1x42:
@clement-pannetier posts already exists for Slovak, I believe. Or am I missing something?

clement-pannetier:
> @clement-pannetier posts already exists for Slovak, I believe. Or am I missing something?

@a1x42, I don't see [posts] inside the `sk.toml` file.

a1x42:
@clement-pannetier lol, classic case of pebkac! it exists only in my fork! ;-)

sdwolfz:
I'm confused, which entries are actually missing from the translation files? (I can submit the Romanian one no problem, just need to know what is needed)

Would it be just...
[...]

PR #471, opened 2020-10-30
Add more Romanian translations

sdwolfz:
### Prerequisites

### Description

Refs: https://github.com/luizdepra/hugo-coder/issues/417

Add missing Romanian translations. For `see_also` it literally translates to "Other from series", I did not want to use the taxonomy name there but it does not sound good without it, compared to English where "Other from" works well without specifying that the next element after that is part of a series. Language quirks...

### DIFF
--- a/i18n/ro.toml
+++ b/i18n/ro.toml
@@ -7,6 +7,12 @@ other = "etichetă"
 [series]
 other = "serie"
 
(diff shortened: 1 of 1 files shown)


luizdepra (commented):
Thank you again. :)

PR #473, opened 2020-10-30
Update variables names to camelCase format for consistency

### LINKED ISSUE #414: [v2] Fix configurations with underline
Hugo uses cameCase for its configuration names. We need to replace `custom_css` and `custom_js` with `customCSS` and `customJS`.

clement-pannetier:
Description
Update configurations with underline. I initially commited 5 months ago but never pushed the code. I cherry-picked the changes and merged with the new configurations.

Issues Resolved
#414

### DIFF
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -1,16 +1,16 @@
-baseurl = "http://www.example.com"
+baseURL = "http://www.example.com"
 title = "johndoe"
 
 theme = "hugo-coder"
 
-languagecode = "en"
-defaultcontentlanguage = "en"
+languageCode = "en"
+defaultContentLanguage = "en"
 
 paginate = 20
 
-pygmentsstyle = "b2"
-pygmentscodefences = true
-pygmentscodefencesguesssyntax = true
+pygmentsStyle = "b2"
+pygmentsCodeFences = true
+pygmentsCodeFencesGuessSyntax = true
 
 disqusShortname = "yourdiscussshortname"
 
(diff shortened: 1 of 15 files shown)


sdwolfz:
:astonished: This is probably going to cause a lot of confusing when upgrading. Maybe for changes like this a Changelog section should be created, just listing the breaking changes should be enough.

@luizdepra would you like me to submit a Readme section listing all the config options and their description? Having the example file is OK but actually documenting them is better!

clement-pannetier:
> This is probably going to cause a lot of confusing when upgrading. Maybe for changes like this a Changelog section should be created, just listing the breaking changes should be enough.

@sdwolfz, there are no breaking changes, hugo is smart enough to see that `hidecolorschemetoggle` is the same as `hideColorSchemeToggle` for example.

Moreover, I made sure that I didn't change the configuration key when using the `isset` function as stated in the doc [here](https://gohugo.io/functions/isset).

> Having the example file is OK but actually documenting them is better!

I documented them when I updated the wiki configurations page. I also changed the configuration keys to camelCase format at the same time.

sdwolfz:
I saw `featured_image` being converted into `featuredImage` in the diff, is this cleanup for an old change?

clement-pannetier:
> I saw `featured_image` being converted into `featuredImage` in the diff, is this cleanup for an old change?

I forgot that I also updated this config, that's the only one that could break old code.

sdwolfz:
OK :shipit: 

Still think that config options should be documented in the repo and not in the wiki, for easy access on search locally. Not a blocker for this PR, and I'm willing to do it myself, just need a couple of :+1:

luizdepra:
@sdwolfz all the configs are present in `exampleSite/config.toml`. This file is something like a reference, and the wiki is a proper documentation. We could add links to the wiki and more comments explaining how each option works.

PR #484, opened 2020-11-15
Unify CDN sources

yujixr:
### Prerequisites

### Description

- Change CDN sources to cdnjs. It may achieve faster load.


yujixr:
I fixed mathjax and katex, so `math.html` uses cdnjs too.

luizdepra:
Thx. :)

rabinadk1:
Sorry, I am commenting after this was merged., but removing defer is not a good idea. And if you had to switch to a common CDN provider, it should have been [jsdelivr](https://www.jsdelivr.com/) as it supports many CDN providers and has the widest coverage range, including China.

PR #486, opened 2020-11-16
Add misssing Polish translations

### LINKED ISSUE #417: [v2] Fix missing i18n strings

0rzech:
Solves #417 for Polish language.


luizdepra (approved):
Thank you. :)

Issue #413, opened 2020-09-28
[v2] Self-hosted dependencies


luizdepra:
Change all JS and CSS dependencies to self-hosting.
Some work was started in an [outdated self-hosted-deps branch](https://github.com/luizdepra/hugo-coder/tree/self-hosting-deps).

We should check if there are conflicts or problems related to the deps' licensing.

plett:
I've grepped for all the external JS and CSS dependencies and looked up what their licenses are.
IANAL, but my interpretation of their licenses is that we would be fine to duplicate all of them in hugo-coder and distribute them. 

For details, baseof uses:
- Google Fonts (Lato, Merriweather, Source Code Pro)
All three of these fonts are under the SIL OFL Open Fonts License
https://www.fontsquirrel.com/license/lato
https://www.fontsquirrel.com/license/merriweather-sans
https://www.fontsquirrel.com/license/source-code-pro
- Fork Awesome 1.1.7
SIL OFL https://forkaweso.me/Fork-Awesome/license/
- Normalize 8.0.1
MIT https://github.com/necolas/normalize.css/blob/master/LICENSE.md
- twemoji
Graphics are CC-BY 4.0 https://github.com/twitter/twemoji/blob/master/LICENSE-GRAPHICS
Project licence looks like MIT, but isn't described as such https://github.com/twitter/twemoji/blob/master/LICENSE

And posts/math uses:
- polyfill
MIT https://github.com/Financial-Times/polyfill-service/blob/master/LICENSE.md
- mathjax
Apache https://github.com/mathjax/MathJax/blob/master/LICENSE
- katex
MIT https://github.com/KaTeX/KaTeX/blob/master/LICENSE

Other optional features:
- utterances
MIT https://github.com/utterance/utterances/blob/master/LICENSE.md
Not suitable for embedding in the theme, it needs to be the same version as the API server https://github.com/utterance/utterances/issues/42

- analytics/plausible
AGPL, but not suitable to embed in the theme, it should come from the Plausible server

- analytics/goatcounter
count.js is ISC, but also not suitable for embedding, it should come from the goatcounter server

@luizdepra How should the next steps be approached? Do you intend v2 of the theme to be self-hosted only with no external JS/CSS resources, or does it need to optional and controlled via a `$.Site.Params` setting?

[...]

PR #488, opened 2020-11-24
twemoji-self-host

pduchnovsky:

### Description

This PR adds self-hosting of twemoji js/assets, part of #413


luizdepra:
As I stated in #413, we're adding tons of files just to serve an optional dependency. It will be hard to maintain and update.

Sorry to make you work in this issue for **maybe** nothing, but I didn't know that tweemoji would add these massive quantity of files.

pduchnovsky:
> As I stated in #413, we're adding tons of files just to serve an optional dependency. It will be hard to maintain and update.
> 
> Sorry to make you work in this issue for **maybe** nothing, but I didn't know that tweemoji would add these massive quantity of files.

Yeah it's a lot of files, maybe it would make sense to self-host the minified js only and leave original reference to their cdn untouched (within the js)

pduchnovsky:
Actually, thinking about it, if the files are not self hosted it also makes no sense to self host the js file neither so I am closing this PR :)

PR #492, opened 2020-12-11
Add support of Cloudflare Web Analytics

yujixr:
### Prerequisites

- [ ] This pull request fixes a bug.
- [x] This pull request adds a feature.
- [ ] This pull request introduces breaking change.

### Description

Add support of Cloudflare Web Analytics.

``` toml
# If you want to use Cloudflare Web Analytics(https://cloudflare.com) for analytics, add this section
[params.cloudflare]
    token = "token"
```

PR #514, opened 2021-01-24
Add abbility to use Google Tag Manager for analytics

xp-vit:
### Prerequisites

### Description

PR adds the ability to use Google Tag Manager code for analytics.

Changes:
1. New template was added for google tag manager.
2. New config option to enable tag manager template added.

### Issues Resolved

### Checklist

#### General

#### Resources

#### Contributors


luizdepra:
Is GTM a replacement to Google Analytics? I think we could go with the same approach used with other analytics options. Creating a configuration and template that is added only when the configuration is set.

Creating an empty template for customization is cool, but imagine if we use this solution for every template we want to customize. It would be a mess.

xp-vit:
GTM is let's say new generation of Google analytics.

With your approach, I totally agree. I will rework it. But I still think this template needs to be still generic custom HTML included to head, not just GTM. Or you prefer an approach for GTM?

luizdepra:
Ok. You could use #492 as an example.

About the header customization, I prefer to encourage people to make a fork or just locally edit things they want to modify. I want to avoid doing the same error I did in footer content. Right now, the footer is a messy template with multiple configurations and combinations, even though everything people asked for could be done by just editing the template.

xp-vit:
@luizdepra I have updated PR. Added dedicated Google Tag Manager template and section on config.toml to enable it by providing google tag manager id

layouts/_default/baseof.html
```diff
@@ -2,6 +2,9 @@
 <html lang="{{ .Site.Language.Lang }}">
 
   <head>
+    {{ if and .Site.Params.googleTagManager .Site.Params.googleTagManager.id }}
+    {{- partial "analytics/googletagmanager" . -}}
+    {{ end }}
```
luizdepra, line 7: I just need to ask you to move these part to the end of the `body` tag, next to the other analytics templates. It doesn't seem like this code really needs to be at the `head` section.

Everything else is fine. :)

xp-vit: I was thinking to do so, but google asks to do it differently:
``` Copy and paste this code as the first item into the <head> of every webpage you want to measure. ```

Should we ignore it?

luizdepra: Sorry for taking too long.

I don't see why moving it would make it not work. Do you mind trying to move the code and testing it?

xp-vit: It works, even when placed lower on the page, however, it's highly recommended to have it in```< head>``` for the following reasons:
 - load as early as possible to make sure that no users are missed
 - A/B testing will not work properly if placed too low

So it's important to keep it in ```<head>```, but we can move it below significant ```<meta>``` tags.

Here are the links with more info:
https://www.bounteous.com/insights/2016/11/22/where-should-google-tag-manager-snippet-be-placed/
https://stackoverflow.com/questions/46064817/how-high-in-the-head-section-can-you-safely-put-the-google-tag-manager-code


do you still feel that we better move it to the bottom of the page?

luizdepra: Hmm... these arguments didn't convince me. 
I prefer to keep things consistent, so we should place it at the end of `body`.

xp-vit: It's changed. Let's merge it :-)

luizdepra:
Nice. Thank you. :)

PR #548, opened 2021-03-27
Add CSP's connect-src directive to config (#547)

### LINKED ISSUE #547: Add CSP's connect-src to configuration and csp.html
I've tried to setup google analytics, but there is an error with missing CSP directive: `connect-src`.

I am getting error in chrome console:
> Refused to connect to 'https://www.google-analytics.com/g/collect?...' because it violates the following Content Security Policy directive: "default-src 'self'". Note that 'connect-src' was not explicitly set, so 'default-src' is used as a fallback.

Checked `/themes/hugo-coder/layouts/partials/csp.html` there is no such attribute `connect-src`

Proposing to add `connect-src` to:
1. new param to configuration as: `.Site.Params.csp.connectsrc`
2. add new `connect-src` directive to layout: `/themes/hugo-coder/layouts/partials/csp.html`

maynagashev:
### Prerequisites

### Description

Adds ability to setup [CSP: connect-src](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/connect-src) directive in config.

### Issues Resolved

#547 Add CSP's connect-src to configuration and csp.html 

#### General
- [x] new param to exampleSite configuration as: `.Site.Params.csp.connectsrc`
- [x] add new `connect-src` directive to layout: `/themes/hugo-coder/layouts/partials/csp.html`


luizdepra:
Ty. :)

PR #549, opened 2021-03-28
Add "buy me a coffee" widget

erickgnavar:
### Prerequisites

- [ X] This pull request adds a feature.

### Description

Add "buy me a coffee" widget

Allow to define the username, color and message for the widget

### Checklist

#### General

A new partial was added to have the widget html, also a new config block was added to specify some variables to be used in the new partial template


erickgnavar:
> Hey! First, thank you for spending your time contributing to the project.
> 
> Well, this feature is cool, but it can be categorized as personal customization. Adding this feature into the theme sets a precedent to adding others donation providers such as ko-fi, patreon, etc. And so, this theme keeps growing bloated with features that other users rarely use. I already made the mistake of accepting multiple comments and analytics systems and I don't want to keep doing it.
> 
> So, I really appreciate your effort and willing to contribute, but I encourage you to keep this kind of change in your fork.

Hi, I totally understand your point :), do you know any other option to add features without having to maintain a fork of the whole theme? It would be nice if there is some kind of "plugin" system so it would be easier to add personal customization.

Thanks for the all your work on this theme 👍

luizdepra:
I think the closest way to a plugin system that I can imagine is to override and use custom templates directly on your site project.
You could create a `layout/` folder in your site repository and move every new and changed files into it, and it should work fine.

PR #555, opened 2021-04-05
Update zh-cn translations and add chinese fonts

lazyyz:
### Description

1. Update zh-cn translations #544 

2. Add chinese fonts


luizdepra (approved):
Nice, thank you. :)

PR #579, opened 2021-06-07
Store theme only when changed from toggle

artspb:
### Prerequisites

### Description

Store theme only when changed from toggle. Otherwise, it never gets changed according to OS settings.

Introduced in ed1c854df2.

#### General

#### Resources

#### Contributors


luizdepra:
Nice. Ty. :)

PR #605, opened 2021-10-06
Add comments for customSCSS file path in config.toml

simonpai:
This commit adds a simple comment on where to put the SCSS files specified in customSCSS setting in config.toml.


luizdepra:
Thanks!

PR #606, opened 2021-10-08
Fix customJS and customCSS feature

simonpai:
### Prerequisites

### Description

In the current state the customJS & customCSS feature doesn't work, because an asset will only be generated to `/public` when `.RelPermalink` is used as mentioned in the [document](https://gohugo.io/hugo-pipes/introduction/#asset-publishing).

This PR fixes the issue by aligning the implementation of including JS asset with the built-in `js/coder.js`, and aligning the handling of CSS with how it handled SCSS. I have tested with Hugo `serve` and `build`.

#### General

#### Contributors

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -123,7 +123,13 @@
     {{ end }}
 
     {{ range .Site.Params.customJS }}
-      <script src="{{ . | relURL }}"></script>
+      {{ if $.Site.IsServer }}
+        {{ $script := resources.Get . }}
+        <script src="{{ $script.RelPermalink }}"></script>
+      {{ else }}
+        {{ $script := resources.Get . | minify | fingerprint }}
+        <script src="{{ $script.RelPermalink }}" integrity="{{ $script.Data.Integrity }}"></script>
+      {{ end }}
     {{ end }}
 
     {{ template "_internal/google_analytics.html" . }}
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -61,7 +61,13 @@
     {{ end }}
 
     {{ range .Site.Params.customCSS }}
-      <link rel="stylesheet" href="{{ . | relURL }}" />
+      {{ if $.Site.IsServer }}
+        {{ $styles := resources.Get . }}
+        <link rel="stylesheet" href="{{ $styles.RelPermalink }}" media="screen">
+      {{ else }}
+        {{ $styles := resources.Get . | minify | fingerprint }}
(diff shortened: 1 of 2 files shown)


luizdepra:
LGTM. Thank you.

layouts/_default/baseof.html
```diff
@@ -61,7 +61,13 @@
     {{ end }}
 
     {{ range .Site.Params.customCSS }}
-      <link rel="stylesheet" href="{{ . | relURL }}" />
+      {{ if $.Site.IsServer }}
+        {{ $styles := resources.Get . }}
+        <link rel="stylesheet" href="{{ $styles.RelPermalink }}" media="screen">
```
gaborbernat, line 66: This commit breaks my website with:

```
ERROR 2022/06/04 09:26:24 render of "taxonomy" failed: "/Users/bgabor8/git/github/bernat-tech/themes/hugo-coder/layouts/_default/baseof.html:66:47": execute of template failed: template: _default/terms.html:66:47: executing "_default/terms.html" at <$styles.RelPermalink>: nil pointer evaluating resource.Resource.RelPermalink
```

luizdepra: Are you using the recommended Hugo's extended version? Only this binary version comes with the Asset Pipeline tooling, and your error seems related.

werat: Before this change I had custom css in `static/css/custom.css` and was refering to it as `customCSS = ["css/custom.css"]`. Now I had to move it to `assets/css/custom.css`. Is this the expected behaviour?

luizdepra: Yes. This is needed because all custom JS and CSS are now processed by the Hugo's Asset Pipeline.

dmohns: Hey, sorry to comment on this ancient issue. I'm trying to update from an ancient version of the theme. But I struggle to get past this commit.

Before, I was using `customCSS` to pull pull in some Google fonts

```toml
 customCSS = [
    "css/custom.css",
    "https://fonts.googleapis.com/css?family=Lato:400,700"
  ]
```

What is the intended workaround here?

I have already done all the other steps as suggested, i.e. moving `custom.css` to `assets` and confirming that I am indeed using extended verison.

luizdepra: I think it will not work with external files. You'll need to download the CSS file from google fonts and add it to your asset folder. Or, you could use @import statements in your custom.css file to load these fonts directly from google fonts.

dmohns: Got it, thanks! For now, I actually just downloaded the entire CSS and put in my assets folder.

PR #619, opened 2021-11-05
Do not set font-size for fontawesome icons in css

### LINKED ISSUE #607: fontawesome size (fa-..x) overwritten
The larger size for the social icons on the home page is overwritten by the `.centered` class.

The example config has `fa-2x` class added, but the CSS is overwritten by the `.centered` class with a fixed **3.2em** size.

See screenshot.

Schermafbeelding 2021-10-08 om 14 43 17
.

alfunx:
### Prerequisites

### Description

Removed the CSS rule which is incompatible with fork-awesome. 
And added `fa-2x` to the example config in order to make the icons sort-of consistent to what we had before.

### Issues Resolved

Closes #607.

### Follow-Up

Should we move the `fa fa-2x` from [the config](https://github.com/luizdepra/hugo-coder/blob/351a501c16c6e1190699ad4d63943297dc159ffc/exampleSite/config.toml#L98) to [layouts/partials/home.html#L23](https://github.com/luizdepra/hugo-coder/blob/master/layouts/partials/home.html#L23), so that you just specify the icon in the config (e.g. just `fa-github` instead of `fa fa-2x fa-github`)? We would lose a bit of "customizability" though…


luizdepra:
LGTM. But, could you please rebuild the conflicting files?
Just run `make release` and `make demo` too.

alfunx:
Done! Rebased, and ran `make release` and `make demo`.

alfunx:
> Should we move the `fa fa-2x` from the config to layouts/partials/home.html#L23, so that you just specify the icon in the config (e.g. just `fa-github` instead of `fa fa-2x fa-github`)? We would lose a bit of "customizability" though…

Do you have any comment regarding this? I guess you missed that part. :slightly_smiling_face:

luizdepra:
> Do you have any comment regarding this? I guess you missed that part. slightly_smiling_face

Oops. Sorry, I missed that part.
Well, I think we can keep the way it is now. In this case, customizability is good.

luizdepra:
LGTM. Thank you. :)

Issue #608, opened 2021-10-13
Missing I18n translations for plural taxonomies


metbril:
In the partial `terms.html` a reference is made to the plural  taxonomy title ('tags', 'categories'), but that isn't yet available in the translation files.

https://github.com/luizdepra/hugo-coder/blob/6581e860bde43960e3b6362001f270cd2e9db178/layouts/partials/terms.html#L9

An option would be to have the full taxonomy name as the translation key and have values for both `one` and `other`. For example:

**en.toml**

```toml
[tags]
one = "tag"
other = "tags"

[categories]
one = "category"
other = "categories"
```

**nl.toml**

```toml
[tags]
one = "label"
other = "labels"

[categories]
one = "categorie"
other = "categoriën"
```

The terms.html partial and possibly other resources need to be updated to use that new key.

luizdepra:
I remember trying to do something like this. I don't know why I got to something like this:

```
[category]
other = "category"

[categories]
other = "categories"
```

Well, your proposal look way better. If someone is interested, please test this solution and open a PR. Thanks.

luizdepra:
I think this one is solved. So, closing.

PR #620, opened 2021-11-12
Plural translation keys

alfunx:
### Prerequisites

### Description

Use plural translation keys for taxonomy titles, as suggested by #608.

### Issues Resolved

Addresses #608 partially. The actual new translations are missing – most languages miss the plural translations (as mentioned in https://github.com/luizdepra/hugo-coder/issues/608#issue-1025635181). For `series` (plural) I added translations in German, French, Italian and Spanish.

### DIFF
--- a/layouts/_default/list.html
+++ b/layouts/_default/list.html
@@ -1,10 +1,9 @@
 {{ define "title" }}
   {{- if eq .Kind "term" -}}
-    {{- i18n .Data.Singular | title -}}
+    {{- i18n .Data.Plural 1 | title -}}
     {{- print ": " -}}
   {{- end -}}
-
-  {{- .Title }} · {{ .Site.Title -}}
+  {{- i18n (lower .Title) | default .Title | title }} · {{ .Site.Title -}}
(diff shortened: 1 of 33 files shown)


layouts/_default/list.html
```diff
@@ -1,10 +1,9 @@
 {{ define "title" }}
   {{- if eq .Kind "term" -}}
-    {{- i18n .Data.Singular | title -}}
+    {{- i18n .Data.Plural 1 | title -}}
```
luizdepra, line 3: Is the`1` intentional? I couldn't find about it in the Hugo's doc.

luizdepra: Got it. It is the value to pick the correct translation. Saw it [here](https://gohugo.io/content-management/multilingual/#query-a-flexible-translation-with-variables).

luizdepra:
Thanks. :)

layouts/_default/list.html
alfunx: Yes exactly. :)

Issue #603, opened 2021-09-22
Code block appears white in dark theme


atz-dev:
Code blocks in markdown appears with a bad contrast and a white background while being on dark mode.  

Here's an example : 
<img width="909" alt="Capture d’écran 2021-09-22 à 22 05 46" src="https://user-images.githubusercontent.com/44492274/134421906-a27dd012-d234-4b14-a9c4-011607c8103e.png">

atz-dev:
changed to monokai, it works as expected. thanks !

(`pygmentsStyle = "monokai"` in config.toml)

PR #596, opened 2021-08-26
Fixed code block background color

LanskovNV:
### Prerequisites

Put an `x` into the box(es) that apply:

- [x] This pull request fixes a bug.
- [ ] This pull request adds a feature.
- [ ] This pull request introduces breaking change.

### Description

This pull request fix code block background color. I made it more contrast at light/dark themes and now it's much more
comfortable to read code on codeblocks

### Checklist

Put an `x` into the box(es) that apply:

#### General

- [x] Describe what changes are being made
- [x] Explain why and how the changes were necessary and implemented respectively
- [ ] Reference issue with `#<ISSUE_NO>` if applicable

#### Resources

- [x] If you have changed any SCSS code, run `make release` to regenerate all CSS files

#### Contributors

- [x] Add yourself to `CONTRIBUTORS.md` if you aren't on it already

PR #630, opened 2021-12-11
Fix bad syntax highlighting color schemes

### LINKED ISSUE #629: Code snippets do not render to dark mode colors
When using the dark theming, if you include a code snippet with ```  it doesn't render particularly dark.

I would be interested in potentially submitting a PR to change this, does anyone have input in terms of how they would like this to look?

This is how it currently renders:
Screen Shot 2021-12-10 at 12 25 10 AM

alfunx:
### Prerequisites

### Description

Some pygment color schemes (like `bw`, which we use as default in the exampleSite) do not set a foreground color explicitly, which makes those themes look bad/illegible when using the dark mode of hugo-coder. Most such themes are light themes (it seems after a quick research), which further explains this phenomenon.

This PR adds a fix for such color schemes by setting the default foreground color to the one used in the light hugo-coder colorscheme. It is overridden by proper syntax highlighting schemes.

### Issues Resolved

Fixes #629, #603, #596.


luizdepra:
LGTM. Thank you. :)

PR #632, opened 2021-12-17
Set `color-scheme` property of `:root` element

alfunx:
### Prerequisites

### Description

By setting the `color-scheme` property of the `:root` element we get matching (dark or light) scrollbars and form controls.

See: https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme


luizdepra:
Thanks!

PR #636, opened 2022-01-17
Add jsdelivr to scriptsrc for Katex

ccviolett:
### Prerequisites

### Description

This pull request add "https://cdn.jsdelivr.net/" to `scriptsrc` in the exampleSite's `config.toml`. 

When run the exampleSite, I try to use Katex but it doesn't work.

I check the Console. It shows that:

So I realize maybe we should add the src of the script to `script-src` in config.toml.

The exampleSite for the new user, like me, who just what to use hugo with hugo-coder theme to write something, should be able to run with no mistake.  

Before this change, the Katex module can't use due to the `Content-Security-Policy`. I found that we put "https://cdn.jsdelivr.net/" to `stylescr` but not to `scriptsrc`. So I did it.

### Issues Resolved

None

### Checklist

#### General

#### Resources

#### Contributors

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -105,3 +105,4 @@
 - [Michael Weiss](https://mweiss.ch)
 - [Simon Pai](https://github.com/simonpai)
 - [Brenton Mallen](https://github.com/brentonmallen1)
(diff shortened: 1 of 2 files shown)


luizdepra:
Thank you. :)

PR #678, opened 2022-05-15
HTML/CSS only tabs

sdwolfz:
### Prerequisites

### Description

Adding shortcodes for creating HTML/CSS only tabs for toggling content. Inspired by:
- https://github.com/rvanhorn/hugo-dynamic-tabs
- https://learn.netlify.app/en/shortcodes/tabs/

Screenshot from 2022-05-15 14-33-08
Screenshot from 2022-05-15 14-33-31
Screenshot from 2022-05-15 14-46-40
Screenshot from 2022-05-15 14-46-54

### Issues Resolved

There are 2 more issues resolved in this PR:
1. Fixing margin top on pre tags. I believe this was a mistake from the 39616beba3da41d4b8b842d9534f9b04f12d330c refactoring. All other elements have both top and bottom margin except this. Visually there is no breaking changes.
2. Added a binding to `0.0.0.0` when running the demo from `make`. This is useful for when you have hugo running in a docker container exposing the `1313` port, otherwise the server can not be reached.

### Technical details

Without requiring JavaScript, there is no way to abstract the DOM/selectors in such a way that I don't need to generate the inline `style` tags. If i use `float: left/right` either the tabs get reversed, or I have to reverse the order of content which seems like it will cause issues for screen readers (not an expert on accessibility, just a guess).  The other solution would be using anchors instead of radio buttons, but those change the URL refs to point to the currently highlighted tab, which is something I did not want from this feature.

Due to all of these, and multiple night of experimentation I concluded that having the generated inline `style` tags was an OK compromise.

### Checklist

#### General

#### Resources

#### Contributors


luizdepra:
This is nice. Thank you.

PR #713, opened 2022-08-11
Make screenreaders do the right thing with heading links.

### LINKED ISSUE #712: hugo-coder breaks Web Content Accessibility Guidelines for links
Currently, hugo-coder breaks Web Content Accessibility Guidelines by having two instances of links with no readable content inside them.  They are the links that produce the heading URLs in render-heading.html and series.html.

This breaks https://www.w3.org/TR/WCAG-TECHS/F89.html

There is a fix for this provided by Font Awesome (and forkawesome) which I will propose in a PR.

simonhollingshead:
### Prerequisites

### Description

Currently, hugo-coder breaks Web Content Accessibility Guidelines by having two instances of links with no readable content inside them.  They are the link icons that produce the heading links.  Because the only content inside them is aria-hidden, the screenreader has nothing to read and will either skip mentioning the presence of the link entirely or will read out the entire URL.

Thankfully, Font Awesome (and by extension Fork Awesome) have an advised solution for this.

https://fontawesome.com/docs/web/dig-deeper/accessibility#semantic-icons

See the part about 'icons that have semantic meaning'.

In this case, we add a hidden element for screenreaders to read, plus a tooltip on the element itself.  Since the element will display or be read, it is generated through i18n so that it can be internationali(s/z)ed.

Note that in the SCSS we have for Fork Awesome, it is called 'sr-only' rather than 'fa-sr-only' but it has identical results.

### Issues Resolved

#712 

### Checklist

#### General

#### Resources

#### Contributors

### DIFF
--- a/layouts/_default/_markup/render-heading.html
+++ b/layouts/_default/_markup/render-heading.html
@@ -1,6 +1,7 @@
 <h{{ .Level }} id="{{ .Anchor | safeURL }}">
   {{ .Text | safeHTML }}
   <a class="heading-link" href="#{{ .Anchor | safeURL }}">
-    <i class="fa fa-link" aria-hidden="true"></i>
+    <i class="fa fa-link" aria-hidden="true" title="{{ i18n "link_to_heading" | default "Link to heading" }}"></i>
+    <span class="sr-only">{{ i18n "link_to_heading" | default "Link to heading" }}</span>
   </a>
 </h{{ .Level }}>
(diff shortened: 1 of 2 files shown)


simonhollingshead:
Please advise me if any changes have to be made to the i18n files (and how we get the correct translations for this string).

luizdepra:
This will work perfectly. Thank you.

Your PR will not break anything about i18n because you provided a default value. But, you could add an entry in every translation file. Something like this:

```
[link_to_heading]
other = "Link to heading"
```

But don't worry. I will add this entry on every translation file later.
Sadly, I will put the English version on almost every file (marking with a `need translation` comment) because I'm unable to translate this sentence to all the languages this theme supports.

PR #719, opened 2022-09-21
Update config.toml

Micro0x00:
avatar image ???? not working

### DIFF
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -1,4 +1,4 @@
-baseURL = "http://www.example.com"
+baseURL = "http://loaclhost:1313"
 title = "johndoe"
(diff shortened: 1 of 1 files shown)


luizdepra:
Hey! You shouldn't need to change the `baseUrl` from an example `config.toml`. Are you experiencing problem to build this theme or your own site?
To test this theme, you can simply use the command `make demo`. But, if you are copying the example `config.toml` to your own project, use `hugo server` to test it. The `hugo` command without any arguments will build your site to production. Using `hugo server`, you will have a simple HTTP Server running your site generated using `localhost:1313` as `baseUrl`, and so ignoring the value configured inside your `config.toml`. You should always configure your `baseUrl` with your production domain name.

PR #755, opened 2023-01-05
Add custom title / translation of categories

marhel:
### Prerequisites

Put an `x` into the box(es) that apply:

- [X] This pull request fixes a bug.
- [X] This pull request adds a feature.
- [ ] This pull request introduces breaking change.

### Description

This PR is similar to #754, but updates the categories to utilize Hugos taxonomy system, instead of the tags .

This PR edits the posts/single layout to use `.GetTerms "categories"` to fetch a richer model of the categories, and the taxonomy/categories partial to use this model, instead of just urlizing the category name, so categories with a custom title, renders correctly on a post. 

This also enables translation of the categories shown on a post, while enabling switching languages on the category page without losing context.

To show this in action, I've added a translation of the "themes" and "syntax" categories in the exampleSite into PT-BR, as "temas" and "sintaxe", respectively, and switched the pt-br page to use the english category ids so they are seen as the same category regardless of translation.

image

Also, this enables customizing/translating the category url, by specifying a slug and/or url in the category frontmatter.

image

If you switch languages here on the category page, you'll see the translated category and corresponding posts, instead of dropping out to the front page like before.

### Issues Resolved

If you wanted to use Hugo taxonomy system for the page categories, only the category id showed up on the post itself. This PR fixes that.

### Checklist

Put an `x` into the box(es) that apply:

#### General

- [X] Describe what changes are being made
- [X] Explain why and how the changes were necessary and implemented respectively
- [ ] Reference issue with `#<ISSUE_NO>` if applicable

#### Resources

- [ ] If you have changed any SCSS code, run `make release` to regenerate all CSS files

#### Contributors

- [ ] Add yourself to `CONTRIBUTORS.md` if you aren't on it already

PR #754, opened 2023-01-04
Add custom title / translation of tags

marhel:
### Prerequisites

Put an `x` into the box(es) that apply:

- [X] This pull request fixes a bug.
- [X] This pull request adds a feature.
- [ ] This pull request introduces breaking change.

### Description

This PR edits the posts/single layout to use `.GetTerms "tags"` to fetch a richer model of the tags, and the taxonomy/tags partial to use this model, instead of just urlizing the tag name, so tags with a custom title, renders correctly on a post. This also enables translation of the tags shown on a post.

To show this in action, I've added a translation of the "shortcodes" tag in the exampleSite into PT-BR, as "Códigos curtos" (my apologies, not a native speaker, this is what Google Translate said when I gave it "short codes"). 

image

The tag still links to /pt-br/tags/shortcodes/ when viewing PT-BR content, and to /tags/shortcodes/ when viewing english content. The same technique can be used to give the shortcodes tag a custom name on the english page as well.

### Issues Resolved

I had a post with a `C#` tag, which on the post tag list linked to /tags/c/ resulting in the "404 page not found" page rendering. Based on [this post on the hugo forums](https://discourse.gohugo.io/t/slash-in-tag-name/38465) I renamed the tag to `CSharp`, and added a page for this tag, where I could give the tag a title of `C#`:

    hugo new tags/CSharp/_index.md

with the content;

    ---
    title: "C#"
    date: 2023-01-04T11:51:36+01:00
    draft: false
    ---

Now, this renders as `C#` on the tags page, linking to /tags/CSharp/ just fine. However, the tag list on the post itself now showed CSharp instead of "C#". This PR fixes that.

### Checklist

Put an `x` into the box(es) that apply:

#### General

- [X] Describe what changes are being made
- [X] Explain why and how the changes were necessary and implemented respectively
- [ ] Reference issue with `#<ISSUE_NO>` if applicable

#### Resources

- [ ] If you have changed any SCSS code, run `make release` to regenerate all CSS files

#### Contributors

- [X] Add yourself to `CONTRIBUTORS.md` if you aren't on it already

PR #756, opened 2023-01-05
Taxonomy authors

marhel:
### Prerequisites

- [X] This pull request fixes a bug.
- [X] This pull request adds a feature.

### Description

This PR is similar to #754 and #755, but updates the authors to utilize Hugos taxonomy system, instead of the tags and categories.

This PR edits the posts/single layout to use .GetTerms "authors" to fetch a richer model of the authors, and the taxonomy/authors partial to use this model, instead of just urlizing the author name, so authors with a custom title from the taxonomy system, renders correctly on a post.

This also enables translation of the authors shown on a post, while enabling switching languages on the author page without losing context.

To show this in action, I've added a translation of the "Hugo Authors" author in the exampleSite into EN and PT-BR, as "Authors of Hugo" and "Autores de Hugo", respectively, and switched  both the EN and PT-BR page for markdown-syntax to use the `authors` frontmatter (instead of `author` like before) so the author is even rendered (most of the pages in the exampleSite use `author` in the frontmatter, which does not show up on the post - I didn't address this)

image

image

If you switch languages here on the author page, you'll see the translated author name and corresponding posts, instead of dropping out to the front page like before.

image

### Issues Resolved

If you wanted to use Hugo taxonomy system for the page authors, only the author ids showed up on the post itself. This PR fixes that.

### Checklist

#### General

- [X] Describe what changes are being made
- [X] Explain why and how the changes were necessary and implemented respectively

#### Resources

#### Contributors

### DIFF
--- a/exampleSite/content/authors/hugo-authors/_index.md
+++ b/exampleSite/content/authors/hugo-authors/_index.md
@@ -1,0 +1,5 @@
++++ 
+draft = false
+date = 2023-01-05T01:15:29+01:00
+title = "Authors of Hugo"
++++
--- a/exampleSite/content/authors/hugo-authors/_index.pt-br.md
+++ b/exampleSite/content/authors/hugo-authors/_index.pt-br.md
@@ -1,0 +1,6 @@
++++ 
+draft = false
+date = 2023-01-05T01:15:52+01:00
+title = "Autores de Hugo"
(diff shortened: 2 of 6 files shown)


luizdepra (approved):
Nice. Thx. :)

Issue #758, opened 2023-01-07
`featuredImage` does not show images in a page bundle


Tempystral:
## Describe the problem:

The page's featured image does not display when using bundled resources. This worked in an older version of hugo-coder but after updating it appears to no longer work.

### Steps to reproduce:

1. Create a page bundle containing an image. I have a few in this bundle.
![image](https://user-images.githubusercontent.com/12647331/211172687-8d204e87-dff0-4370-b0cc-e2855ee22dfb.png)
2. Set that image as the `featuredImage`

### Observed Results:
The image does not display:
![image](https://user-images.githubusercontent.com/12647331/211172662-d77e71d4-9efe-4c74-b7d8-4a4148922b6c.png)

I believe this is because of the ` | relURL` pipe in the `posts/single.html` partial. If I remove that line and just display `.Params.featuredImage` it's fine:

![image](https://user-images.githubusercontent.com/12647331/211172669-82e7b4a7-8790-45fa-ba4b-9e8bc591c876.png)

I imagine having this relURL is probably useful for other use cases but it would be nice if Hugo supported bundled images in the header for easy asset organization.

PR #759, opened 2023-01-07
Get featuredImage from assets/ dir or page bundles

Tempystral:
### Prerequisites

### Description

This change allows the `featuredImage` front matter param to reference images in both the `assets/` directory as well as in a [page bundle](https://gohugo.io/content-management/page-bundles/)

### Issues Resolved

https://github.com/luizdepra/hugo-coder/issues/758

### Checklist

#### General

#### Contributors

### DIFF
--- a/layouts/posts/single.html
+++ b/layouts/posts/single.html
@@ -33,7 +33,7 @@
 
       <div class="post-content">
         {{ if .Params.featuredImage }}
-          <img src="{{ .Params.featuredImage | relURL }}" alt="Featured image"/>
+          <img src="{{ or (.Resources.GetMatch .Params.featuredImage) (resources.Get .Params.featuredImage).RelPermalink (.Params.featuredImage | relURL) }}" alt="Featured image"/>
         {{ end }}
         {{ .Content }}
       </div>
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -123,3 +123,4 @@
 - [Brian Lachniet](https://github.com/blachniet)
 - [ShortArrow](https://github.com/ShortArrow)
 - [Martin Hellspong](https://github.com/marhel)
+- [Tempystral](https://github.com/tempystral)


layouts/posts/single.html
```diff
@@ -33,7 +33,7 @@ <h1 class="title">
 
       <div class="post-content">
         {{ if .Params.featuredImage }}
-          <img src="{{ .Params.featuredImage | relURL }}" alt="Featured image"/>
+          <img src="{{ or (.Resources.GetMatch .Params.featuredImage) (resources.Get .Params.featuredImage).RelPermalink (.Params.featuredImage | relURL) }}" alt="Featured image"/>
```
luizdepra, line 36: Hmm... I don't know. This looks like overly complicated. `GetMatch` could be a problem, because of its pattern matching it could match with a similarly named file. And, `resources.Get` will try to find the image inside the `assets` folder, which is not related.

I see two possible solutions for your problem.
1. Just removing `relURL` from the original line, so you could use relative path to the page and not to the base URL. (It makes sense, but we'll need to remove `relUr` from the favicon and avatar code too.)
2. Use a full path in `featureImage`  such as `posts/page-bundle/images/something.jpg`.

I don't mind going with the option (1), but people will complain about the breaking change.

Tempystral: I suppose I could use a full path in the featured image, it would just be nice to be able to use a relative path like you can with page resources. I definitely don't want to introduce a breaking change.

I don't know that it's too complicated. The goal is to enable the option for bundled images with a relative path, and whatever evaluates truthy first is picked.

That said, I've done some more testing and realized that if I define a page resource with the same path, the image doesn't render correctly, so I suppose I can see the issue there.

I think I will just use an absolute path, or override the `single.html` template with my own.

luizdepra: Yeah, currently this is a practical option. But let me better check other solution this weekend. I think we could create a custom template to render image links and abstract this code. I'm also thinking that using the asset pipeline for images could be a good idea, but I need to understand it better first.

Tempystral: Sure thing. I'm relatively new to using the asset pipeline as well. If you'd like I can draft something in this PR or we can close it and I'll let you do your thing.

PR #760, opened 2023-01-08
fix: correct regression with code highlighting in dark mode

robertwtucker:
### Prerequisites

### Description

This PR reverts a regression error in which the foreground color specified for highlighted code is almost indistinguishable from the white background used in dark mode.

This was originally fixed in 29a7a94 / #630 by assigning `color: $fg-color` from the light scheme variables (resolves to `#212121`). In 476429d, the assignment was changed to `$fg-color-dark` (resolves to `#dadada`) but this resulted in a color that is too close to the white of the code highlight to be read easily. See image:

SCR-20230107-rqc

This fix assigns `$bg-color-dark` which resolves to `#212121` (same as the original fix). With this change applied, the code is now legible:

SCR-20230107-rol

### Issues Resolved

N/A

### Checklist

#### General

#### Resources

#### Contributors


luizdepra:
I will accept this, but I fear that using a different option of code highlight colors could result in the same problem. Let's see.

luizdepra (approved):

PR #777, opened 2023-02-25
Make H1 sizes consistent between different page types

simonhollingshead:
### Prerequisites

### Description

Right now, "article header h1" gets a special larger heading than any other heading.  For consistency's sake, all headings should have the same size. 

This change does not override all H1s otherwise headings generated halfway down a page with a single '#' in markdown would look weirdly large.  It also does not change all pages to use the article tag, since the specification states an article must be independently excerpt-able content, and I would not say that any flavour text above a page listing is, by itself, an article (e.g. the stuff that you might put as content for list.html probably doesn't stand alone as an article).

Instead, this chooses to just resize anything which meets the semantic element of "header h1" (and it adds the semantic element in places where a header should really be acknowledged as a header, such as on some of the partials).

### Issues Resolved

https://github.com/luizdepra/hugo-coder/discussions/776

### Checklist

#### General

#### Resources

#### Contributors


luizdepra (approved):
Looks good! Thank you.

PR #794, opened 2023-04-06
Be stricter about what counts as an external link.

simonhollingshead:
### Prerequisites

### Description

The current matching of adding the external-link class has both false positives and false negatives.  This attempts to be slightly stricter about what is an external link.

Before, things like http://foo.com correctly were identified as external.  However:

* A relative link to "http-specification-2.0.pdf" would be considered external because it starts with 'http'
* HTTP://EXAMPLE.COM is considered internal because 'HTTP' != 'http'
* //example.com (i.e. a protocol-relative destination) is considered internal because it does not start with 'http'.

This resolves those matters by strictly checking for the '://' in the forcibly-lowercased version of the destination, and for URLs starting with two slashes.

### Checklist

#### General

#### Resources

#### Contributors

### DIFF
--- a/layouts/_default/_markup/render-link.html
+++ b/layouts/_default/_markup/render-link.html
@@ -1,1 +1,1 @@
-<a href="{{ .Destination | safeURL }}"{{ with .Title}} title="{{ . }}"{{ end }} {{ if or (or (strings.HasPrefix (lower .Destination) "http://") (strings.HasPrefix (lower .Destination) "https://")) ( strings.HasPrefix .Destination "//" ) }} class="external-link" target="_blank" rel="noopener"{{ end }}>{{ .Text | safeHTML }}</a>
+<a href="{{ .Destination | safeURL }}"{{ with .Title}} title="{{ . }}"{{ end }} {{ if findRE "^(https?:)?//" (lower .Destination) }} class="external-link" target="_blank" rel="noopener"{{ end }}>{{ .Text | safeHTML }}</a>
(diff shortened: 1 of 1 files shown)


layouts/_default/_markup/render-link.html
```diff
@@ -1 +1 @@
-<a href="{{ .Destination | safeURL }}"{{ with .Title}} title="{{ . }}"{{ end }} {{ if strings.HasPrefix .Destination "http" }} class="external-link" target="_blank" rel="noopener"{{ end }}>{{ .Text | safeHTML }}</a>
\ No newline at end of file
+<a href="{{ .Destination | safeURL }}"{{ with .Title}} title="{{ . }}"{{ end }} {{ if or (or (strings.HasPrefix (lower .Destination) "http://") (strings.HasPrefix (lower .Destination) "https://")) ( strings.HasPrefix .Destination "//" ) }} class="external-link" target="_blank" rel="noopener"{{ end }}>{{ .Text | safeHTML }}</a>
```
luizdepra, line 1: I think we could simplify here using `findRE`.

Maybe something like:
```
{{ if findRE(`^(http)?s?:?//` (lower .Destination)) }}
```

simonhollingshead: I was originally avoiding using a regex for performance reasons but now I realise that the performance impact is absolutely negligible.

luizdepra (approved):
Thanks again. :)

PR #802, opened 2023-04-24
Implement Mastodon comments

retiolus:
### Prerequisites

- [ x] This pull request adds a feature.

### Description

Implements Mastodon comments with Carl Schwan solution: https://carlschwan.eu/2020/12/29/adding-comments-to-your-static-blog-with-mastodon/

This allow to link post with a Mastodon toot, comments on the Mastodon toot will be displayed on the post footer.

### Checklist

#### General

- [ x] Describe what changes are being made
- [ x] Explain why and how the changes were necessary and implemented respectively

#### Resources

#### Contributors


luizdepra:
> Now the comment system does work and comments are charging, but I don't know why I can't manage to use local purify.min.js...

What do you mean? Do you want to make the `purity.js` file self-hosted?
If so, you need to add it or in the `static` or in the `assets` folder and change the `script` line that loads it.

But, I recommend you to load the script from a CDN. We don't want to add another JS file into the project that is only used if the user enables an optional feature.

luizdepra:
Well, all included JS and CSS files are loaded every time (page refresh/navigation), regardless of where it is hosted. If the file is served by a CDN or by your host of choice, your browser will always download the file. Ok, not always because the browser has its cache.

People like to host every dependency on the same host of their page because they want to avoid depending on a third-party provider. So, if external providers go down, their site will not suffer from it.

There are ways to use custom configs to set a path to load a local version of any dependency. But, I think it is too much effort to something so simple, and bloating this theme with one more configuration. Notice that your site's comment system will be depending on an external service (Mastodon) anyway.

luizdepra:
I forgot about it. xD
Later I'll take a look again and probably change a thing or two, in a separated branch.

PR #834, opened 2023-08-05
add support for extra class in menu links

tobiasehlert:
### Prerequisites

### Description

Adds possibility to add extra class(es) to navigation-link items in the menu.

### Issues Resolved

_n/a_

### Checklist

#### General

#### Resources

#### Contributors

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -12,7 +12,7 @@
         {{ with .Site.Menus.main}}
           {{ range sort . }}
             <li class="navigation-item">
-              <a class="navigation-link" href="{{ index (apply (slice .URL) (.Params.urlFunc | default "relLangURL") ".") 0 }}">{{ .Name }}</a>
+              <a class="navigation-link {{ .Params.class }}" href="{{ index (apply (slice .URL) (.Params.urlFunc | default "relLangURL") ".") 0 }}">{{ .Name }}</a>
             </li>
           {{ end }}
         {{ end }}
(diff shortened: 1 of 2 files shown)


tobiasehlert:
Would maybe even be better if we also add support for "active" page support as in example:
https://gohugo.io/templates/menu-templates/#example

luizdepra:
Sorry for taking so long to reply here. Well...

Sure. We could use this example as a base to add multi-level menus.

If you're interested, feel free to send a PR changing this code.

layouts/partials/header.html
```diff
@@ -12,7 +12,7 @@
         {{ with .Site.Menus.main}}
           {{ range sort . }}
             <li class="navigation-item">
-              <a class="navigation-link" href="{{ index (apply (slice .URL) (.Params.urlFunc | default "relLangURL") ".") 0 }}">{{ .Name }}</a>
+              <a class="navigation-link {{ .Params.class }}" href="{{ index (apply (slice .URL) (.Params.urlFunc | default "relLangURL") ".") 0 }}">{{ .Name }}</a>
```
luizdepra, line 15: Let me just ask for one more thing. Could you add an entry in the `docs` folder about this new configuration parameter?

tobiasehlert: @luizdepra, I've added a entry in the docs with commit 28d4538.
Is that good enough or you want me to extend that even further?

luizdepra: LGTM. Ty. :)

PR #836, opened 2023-08-12
remove Song Ti from font list for backward compatibility in Japanese

mqcmd196:
### Prerequisites

### Description

https://github.com/luizdepra/hugo-coder/pull/555 added the new font for Chinese, but 宋体 changed the default Japanese font. This PR removes it.

I don't know whether this is the correct way to fix...

### Issues Resolved

Before this PR
image

After this PR
image

I think the font before this PR is quite unnatural for current Japanese people.

### Checklist

#### General

#### Resources

#### Contributors


luizdepra:
> I don't know whether this is the correct way to fix...

Me too. I need to research a bit about it, but I think we'll need to come up with a better font stack. If anyone could help with that, I would appreciate.

Well, removing the Chinese font will probably change someone's project. So, I advise you to override the variables file and change the font stack, or just keep a fork with the change. We have to find a good way to handle both cases, if possible.

mqcmd196:
Thank you for your advice.

What is the correct way to do 
> I advise you to override the variables file and change the font stack

?

luizdepra:
You'll need to make a copy of the [_variables.scss](https://github.com/luizdepra/hugo-coder/blob/main/assets/scss/_variables.scss) file and make the font change. The copy should be placed in the same folder structure as the theme. So, add it to `assets/scss/_variables.scss`. So, Hugo is clever enough to understand that you are overriding only this file.

PR #840, opened 2023-08-16
Fixing problem with incorrect style of utterances comment section

piotrek-k:
### Prerequisites

### Description

This PR fixes a bug that causes inconsistencies between the themes used by the website and the utterances comments section.

If the website theme is set automatically and the user is in dark mode, utterances will use light mode.

This is because the `colorscheme` parameter in localStorage is not set until the button to switch themes is used, and if this parameter is not set, utterances displays the light theme by default.

Steps to reproduce:
1. Use dark mode on your system
2. Load a website with "utterances" comment section enabled
3. See that although your website is in dark mode, utterances uses light mode

### Issues Resolved

None


luizdepra (approved):
Nice. Thanks. :)

PR #860, opened 2023-11-21
Fix umami analytics script name

### LINKED ISSUE #859: Umami analytics feature is broken
## Describe the problem:

Even though my site is deployed and getting visits (from myself), the umami dashboard tells me there is no activity at all. This is almost certainly due to the tracking script being renamed from "umami.js" to "script.js". PR incoming.

iron3oxide:
Apparently umami renamed their script which silently broke the feature in hugo-coder.

closes #859

### Prerequisites

### Description

It fixes the umami analytics feature.

### Issues Resolved

#859 

### Checklist

#### General

#### Contributors

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -135,3 +135,5 @@
 - [Kemal Akkoyun](https://github.com/kakkoyun)
 - [Igetin](https://github.com/Igetin)
 - [Kirill Che.](https://github.com/g4s8)
+- [iron3oxide](https://github.com/iron3oxide)
+  
--- a/docs/analytics/umami.md
+++ b/docs/analytics/umami.md
@@ -3,5 +3,7 @@
 ```toml
 [params.umami]
     siteID = "ABCDE"
-    serverURL = "analytics.example.com"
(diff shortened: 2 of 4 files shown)


iron3oxide:
On second thought, maybe we should replace "serverURL" with "scriptURL"? I found the former to be somewhat confusing as it is neither explained with a comment nor featured in the info umami gives you to integrate websites. Plus it would obviously safeguard against future renaming of the script.

iron3oxide:
I'll be glad to adapt the docs, however I don't quite understand yet which files need to be changed in order to properly rename a config parameter. E.g. https://github.com/luizdepra/hugo-coder/blob/c4665043f9957b7ed9b5a07c7074bb520563a5ba/layouts/_default/baseof.html#L91
and https://github.com/luizdepra/hugo-coder/blob/c4665043f9957b7ed9b5a07c7074bb520563a5ba/layouts/partials/analytics/matomo.html#L7
come up when searching for `matomo.serverURL` in the repo.

The latter seems to need a change either way since it uses  a separate server url and (old) script name, but are those the only files that would have to be adapted?

luizdepra (approved):
Nice. Thank you. :)

PR #870, opened 2023-12-05
Allow users to disable default theme-switching Javascript files

### LINKED ISSUE #831: Add an option to disable JavaScript
## Describe the enhancement

There should be a configuration option to disable JavaScript altogether. When the value of this setting is `true`, no `<script>` tags or `<link>` tags pointing to JS resources should be included on any page. This should also imply disabling any of the theme features that depend on JavaScript (like the light/dark toggle button).

Currently one needs to override [the `baseof.html` template](https://github.com/luizdepra/hugo-coder/blob/4a6b8fd3551842e89dd15f8181056fe48a3b04f7/layouts/_default/baseof.html#L36-L42) in order to remove `coder.js` from the pages.

mmalik23:
### Prerequisites

### Description

This PR enables users to use the theme independently of JavaScript, requiring only minimal changes. I've introduced a `disableDefaultJsScripts` flag, rather than a `disableJS` toggle, to avoid implying zero JavaScript. This approach keeps separately configured JavaScript features like comments and analytics functional, ensuring they are logically consistent with the `disableDefaultJsScripts` configuration. 

Moreover, when disableDefaultJsScripts is enabled, hideColorSchemeToggle must be set to true, or else the app will fail with an error message. I felt that this is a misconfiguration and alerting the user would provide a better overall user experience ( I would not want somebody spending time trying to figure out why the theme is not switching)

### Issues Resolved

#831

### Checklist

#### General

#### Resources

#### Contributors

### DIFF
--- a/docs/configurations.md
+++ b/docs/configurations.md
@@ -111,8 +111,8 @@ These are all the parameters used by `hugo-coder` theme.
 | description                   | string | Yes      | Description of the site.                         |                                  | `"John Doe's personal website"`                  |
 | keywords                      | string | Yes      | Site keywords.                                   |                                  | `"blog,developer,personal"`                      |
 | avatarURL                     | string | No       | Photo of the author.                             |                                  | `"images/avatar.jpg"`                            |
-| gravatar                      | string | No       | Gravatar photo of the author                     |                                  | `"john.doe@example.com"`                       |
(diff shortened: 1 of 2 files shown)


luizdepra:
I will accept this, but I think I will need to rework the way the color scheme code works. I should not be necessary to have `coder.js` when not using the toggle, but I think it will break if not.

PR #882, opened 2024-01-19
Swap from ForkAwesome 1.2.0 to FontAwesome 6.5.1.

### LINKED ISSUE #858: Fork-Awesome is deprecated, please consider alternatives
## Describe the problem:

I've been watching Fork-Awesome for a while, a bit frustrated by bugs, missing icons and the lack of new releases. I even recompiled it myself to use a newer version of it, only to discover [more bugs](https://github.com/ForkAwesome/Fork-Awesome/issues/421).

Turns out the Fork-Awesome consider the project deprecated (or even "failed" in some ways), and plan to archive it in 2024. See https://github.com/ForkAwesome/Fork-Awesome/blob/master/README.md and https://github.com/ForkAwesome/Fork-Awesome/blob/master/posts/2023-04-10_forkawesomeIntrospection/forkawesomeIntrospection.md.

I think hugo-coder should start looking for alternatives, which may include switching back to Font Awesome.

simonhollingshead:
### Prerequisites

### Description

This change allows FontAwesome 6.5.1 to be used as a drop-in replacement for ForkAwesome 1.2.0.

The easiest way to see that my change has worked practically-invisibly is that the home page of the demo site now shows Medium with its new three-circular-blobs logo instead of the old capital-M logo, despite the fact I have not changed anything about the hugo configuration for the demo site.

Points of note here:

- Using FontAwesome in SCSS is different.  Therefore the ONLY breaking change here is for people who have made changes to the theme and written "@include fa-icon" in their SCSS.  Everything that can happen in content (e.g. class names) are translated by the v4-shims.
- All files here are from https://fontawesome.com/download ('Free for Web').
  - The contents of the /scss folder are in assets/scss/font-awesome.
  - The contents of /webfonts are in static/fonts (except fa-v4compatibility because that's not needed - v4-shims does all the work we need).
  - /LICENSE.txt is added to both directories to ensure it is clear both are part of the FA-licensed code.
- In ForkAwesome, the path to the font was constructed in _path.css suffixing $fa-version, so the preload needed to have the version at the end.  Since the default behaviour of FontAwesome is to not include a version suffix, the preloads no longer have one either.  This is a necessary change because the HTML spec (section 4.6.6.19) explicitly says that the entire URL [i.e. including query and fragment] is in the key used to determine whether an element in the preload cache can be consumed.

### Issues Resolved

#858 

### Checklist

#### General

#### Resources

#### Contributors


simonhollingshead:
I just wanted to add a second copy of the LICENSE.txt so that both directories with FA code contain the FA license.

sammy44nts:
Thank you for the great work!
Is it possible to fetch FontAwesome 6.5.1 from a CDN instead of importing it in hugo-coder own CSS ?

For SEO purpose, I'm planning to add an issue about uglifying JS and use separated CSS instead of one big CSS in order to improve first paint.

simonhollingshead:
It would be able to be done, but absolutely not as part of this specific PR.  You're right that it would require splitting the CSS out so that it could be overridden.  Otherwise, we'd be forcing external website dependencies on all users of the template which would be a big behaviour change which wasn't discussed in the original issue.

If you made the appropriate changes to split the CSS into multiple files, then replacing the fontawesome CSS with the CDN links at https://cdnjs.com/libraries/font-awesome would 'just work' - it'd automatically use the font files also hosted there.

luizdepra:
Thank you!

PR #887, opened 2024-01-30
Swap all use of FontAwesome to the v5+ syntax.

simonhollingshead:
### Prerequisites

### Description

This change is the second half of #882, designed to remove use of the v4 syntax from the codebase.

This is not removing the v4 shims as users may require it for their own content, but we don't need the template to rely on the old syntax!

### Issues Resolved

I promised this cleanup in #858 

### Checklist

#### General

#### Resources

#### Contributors


luizdepra:
Looks good. Thank you again. :D

PR #900, opened 2024-03-28
feat: introduce vercel web analytics

xiaotianxt:


layouts/partials/analytics/vercel.html
```diff
@@ -0,0 +1,4 @@
+<script>
+    window.si = window.si || function () { (window.siq = window.siq || []).push(arguments); };
+</script>
+<script defer src="/_vercel/speed-insights/script.js"></script>
```
luizdepra, line 4: Is this path exclusive to Vercel hosting, right? Don't we need to add this file into the project, or it is automatic embedded when using Vercel?

Just a problem here. Using absolute path (starting with `/`) will not work when hosting when the base path is something like this: `mysite.com/blog/`. It will try to find the Vercel script under `mysite.com` instead of `mysite.com/blog/`.

Could you please fix that first?

luizdepra (approved):
Nice. Thank you. :)

PR #911, opened 2024-05-05
Add Copy Button

essesoul:
### Prerequisites

### Description

Click on the button to copy the content of the code block

One-click copying of the content of a block of code is achieved by creating the files `static/css/clipboard.css` and `static/js/clipboard.js`

Users can now quickly copy the contents of a block by clicking the button in the upper right corner of the block.

### Checklist

#### General

#### Resources

#### Contributors

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -3,6 +3,7 @@
 - [Chip Senkbeil](https://github.com/chipsenkbeil)
 - [Dale Noe](https://github.com/dalenoe)
 - [Gabor Nagy](https://github.com/Aigeruth)
+- [Gary Chan](https://github.com/essesoul)
 - [Harry Khanna](https://github.com/hkhanna)
 - [Ihor Dvoretskyi](https://github.com/idvoretskyi)
 - [Jacob Wood](https://github.com/jacoblukewood)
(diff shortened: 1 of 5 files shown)


luizdepra:
Nice, thank you.
But I will need to make some modifications like move the CSS code to SCSS, and overriding the code template to add the button instead of using JS to include it.

Yunglow22 (approved):

septs (commented):
i'm not a reviewer, but i think this is not elegant

1. i think it would be better to use `js.Build` preprocessing

```
{{-
$defines := dict
    "css_url" (resources.Get "js/clipboard.css").Permalink
-}}
{{- with resources.Get "js/clipboard.ts" (dict "defines" $defines) | js.Build | minify | fingerprint -}}
<script src="{{ .Permalink }}"></script>
{{- end -}}
```

2. use `layouts/_default/_markup/render-codeblock.html` hook, set page variable

```
{{- .Page.Store.Set "hasCodeblock" true -}}
```

see https://gohugo.io/render-hooks/code-blocks/

4. use javascript dynamic loading stylesheet

```typescript
declare const css_url: string

const stylesheet = document.createElement("link")
stylesheet.rel = "stylesheet"
stylesheet.href = css_url
document.head.appendChild(stylesheet)
```

jemus42:
I was just about to research how to add a code copy button when I found this --- is this still being considered?

pduchnovsky:
`clipboard.writeText(codeBlock.innerText)`
must be
`clipboard.writeText(codeBlock.textContent)` instead

innerText can add extra line breaks or modify spacing due to CSS/rendering.
textContent preserves the exact formatting in the DOM — better for code blocks.

PR #931, opened 2024-09-23
Hide fediverse creator when unused, and allow a globally-set value.

### LINKED ISSUE #930: Fediverse creator tag should fall back to a site-level param if the value is not in the page params, and should not be present if unset.
## Describe the enhancement

In a relatively recent change, the fediverse creator tag was added (#920).  However, this uses .Params with no default to .Site.Params.

If I am writing a lot of articles on my site, it stands to reason that I should be able to set my fediverse information once in my hugo.toml and have it propagate through onto all pages.  As such, this should really support a default value.

In addition, having an empty meta tag on every single page when this value is not set is not ideal.  In the case where no value is set, the entire tag should be elided.

simonhollingshead:
### Prerequisites

### Description

The new-ish changes to support fediverse creator tags seemed to introduce two undesirable (to me) features.

1. People who didn't use the feature got an empty meta tag added to every page with no value.
2. People who own their site and are the only author of content to it needed to add the value on every single page to get it to function.

This change allows for a default value to be set, and removes the tag if it is valueless.

### Issues Resolved

#930 

### Checklist

#### General

#### Resources

#### Contributors


luizdepra:
Thank you. :D

PR #965, opened 2025-04-23
Hide the date if `showthedate` is false

SvenHaedrich:
### Prerequisites

- [X] This pull request introduces breaking change.

### Description

This PR hides the date in the list of blog entries depending on the `.Params.showthedate` settings.

### Checklist

#### General

- [ ] Describe what changes are being made

#### Resources

#### Contributors

### DIFF
--- a/layouts/partials/list.html
+++ b/layouts/partials/list.html
@@ -14,7 +14,7 @@
   <ul>
     {{ range .Paginator.Pages }}
     <li>
-      {{ if .Params.showthedate }}
+      {{ if not (.Site.Params.hidedates) }}
       <span class="date">{{ .Date | time.Format (.Site.Params.dateFormat | default "January 2, 2006" ) }}</span>
       {{ end }}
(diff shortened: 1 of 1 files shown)


layouts/partials/list.html
```diff
@@ -14,7 +14,9 @@ <h1 class="title">
   <ul>
     {{ range .Paginator.Pages }}
     <li>
+      {{ if .Params.showthedate }}
```
dbryar, line 17: I'd use a negative assertion so a missing param is defaulting to show dates, i.e.
- use `.Params.hidedates` instead of `.Params.showthedate`

SvenHaedrich: Using negative logic makes this a non-breaking change. Also, I made `hidedates` explicitly a `Site` parameter.

luizdepra:
Hey. Thank you for using and contributing.

I'm not accepting this change because I'm trying to avoid adding more configurations options that could be a template customization.

